Issue #4, opened 2025-01-20
[FEATURE] Cache locally 'failed' transactions + Display on Wallet


andrewnguyen22:
### Describe the feature you'd like
1. Track locally (from CLI or WebWallet) submitted transactions to see if they fail in the mempool at any phase (when it first arrives - or later when the mempool is 'rechecked')
2. Display the failed transaction and the`lib.ErrorI` in wallet under 'recent transactions'

### Motivation
When a transaction fails, a user is left wondering why without parsing the logs.

### Additional context
- If a transaction is included in a block - remove it from the cache
- Just because a transaction is valid when it first arrives to the mempool, doesn't mean it stays valid
- Minimal code and simplicity is priority over robustness

#### Required Tags
- Priority: Medium
- Module: Controller, WebWallet

PR #20, opened 2025-01-23
Issue #4 - Failed Transaction cache + display

rem1niscence:
## Description
Implemented a failure transactions cache on the mempool, to check on failed transactions, added a server endpoint for it and display those transactions in the wallet.

## Related Issues
Closes: #4 

## Changes Made
- Adds a failed transactions cache to the mempool, which continuously monitors for failed transactions and their accompanied error
- Implemented a `FailedTransactions` endpoint to be consumed by the wallet at the moment
- Display failed transactions on the wallet under the "recent" transactions table
- Fix a minor bug on wallet number formatting which could allow numbers in certain cases

## Checklist

Although this is mentioned on the issue, such check is not implemented, my reasoning is that once a transaction has been marked as false, either right away or after a recheck of the mempool, it is completely remove, so there's no need to check for the block transactions as the failed ones will never be submitted, feel free to correct me if I'm wrong

Also, a comment could be made o why I made a separate struct type instead of converting a `map[string]FailedTx` into it's own type like:

```go
type FailedTxCache map[string]*FailedTx
```

And that's because I then wouldn't have complete control of which items could be added, such as only allowing specific message types or unmarshalling the tx bytes into a `lib.Transaction` by having a struct I can fully control the access and perform safety checks on the input.

### DIFF
--- a/.docs/CHANGELOG.md
+++ b/.docs/CHANGELOG.md
@@ -1,11 +1,18 @@
 # CHANGELOG
 
+## [Unrelease] - 1/23/24
+- Add Failure transactions cache
+- Add Simple paginator query with no additional params
+- Implement Failure transactions endpoint
+- Display failed user transactions in the wallet
+- Fix Wallet Bug on number input
+
 ## [Unreleased] - 1/21/2024
 - Set up prettier for both Wallet and Explorer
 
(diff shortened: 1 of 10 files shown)


andrewnguyen22 (approved):
Awesome job

Issue #12, opened 2025-01-20
[FEATURE] Add `max` option to send the entire address balance


andrewnguyen22:
### Describe the feature you'd like
In the web-wallet, add a `max` or `all` option to send the entire address balance

Should be implemented for all transaction types with an `amount` field

<img width="545" alt="Image" src="https://github.com/user-attachments/assets/e77e9c30-3457-4eff-b565-b81fff1ad499" />

### Motivation
Would allow the user to easily empty a wallet which is a typical behavior

#### Required Tags
- Priority: Low
- Module: Wallet

PR #28, opened 2025-01-28
Issue-#12: Max amount form button

rem1niscence:
## Description
Add max amount form button

## Related Issues
Closes: #12 

## Changes Made
- Added max amount form button
- Changed `renderForm` function to manage the state internally as an object 

## Checklist

## Additional Notes
The `renderForm` function may seem to make the object a bit more complex and while its true, it is necessary to allow for react to manage its fields, because the`renderFom` may render different forms with different amount of fields, making use of `useState` per field leads to an `"Error: Rendered more hooks than during the previous render.` if you change to any other form, this new approach allows forms to be of any size without disrupting react rendering and allow for more complex interactions and checks for the future against any kind of form.

image
(as it can be seem there's a small issue on  `formatNumber` when dealing with decimals, that can be tackled on another issue)

### DIFF
--- a/.docs/CHANGELOG.md
+++ b/.docs/CHANGELOG.md
@@ -1,6 +1,9 @@
 # CHANGELOG
 
 ## [Unreleased] - 1/28/2024
+- Add max amount form button
+
+## [Unreleased] - 1/28/2024
 - Fix UI transactions table Issue
 
 ## [Unreleased] - 1/24/2024
--- a/cmd/rpc/web/wallet/components/account.jsx
+++ b/cmd/rpc/web/wallet/components/account.jsx
@@ -302,23 +302,67 @@ export default function Accounts({ keygroup, account, validator }) {
   }
 
   // renderForm() returns a form input group for the transaction execution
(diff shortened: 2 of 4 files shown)


rem1niscence:
I'm finishing another PR, can do as soon as I'm done with this one

andrewnguyen22 (changes requested):
Looks good, but anyway to keep the input field spacing consistent even with the max button?

rem1niscence:
> Looks good, but anyway to keep the input field spacing consistent even with the max button?
I think it can be done but it will look cramped IMO, as the balance text will have no spacing with the next input, you want that?

andrewnguyen22:
> > Looks good, but anyway to keep the input field spacing consistent even with the max button?
> 
> I think it can be done but it will look cramped IMO, as the balance text will have no spacing with the next input, you want that?

I mean it all be the same spacing as you have for the balance text

andrewnguyen22 (approved):
Disregard above comment - looks good as is

Issue #35, opened 2025-01-30
[FEATURE] Remove zero balance pools from state


andrewnguyen22:
### Describe the feature you'd like
A pool with zero balance should be removed from the state.

Example: 
```json
{
      "id": 1,
      "amount": 0
}
```

### Motivation
This feature would ensure no unnecessary bloating of the state structure from pools.

### Additional context
This modification is consensus breaking and should happen pre-mainnet

### Possible implementation
(Optional) If you have ideas for how this feature could be implemented, share them here.

#### Required Tags
- Priority: High
- Module: FSM (Finite State Machine)

PR #36, opened 2025-01-31
feat: not return 0 amount pools in state

pablocampogo:
## Description
- "Delete" accounts and pools that have 0 amount

## Related Issues

Closes: #35   

## Changes Made
- [x] Change `SetAccount` and `SetPool` to delete pools and accounts with amounts set to 0.

## Checklist


andrewnguyen22 (changes requested):
Hey @pablocampogo thanks for the super clean PR. Very easy to review!

Your solution does not satisfy the issue however, because it's just making zero pools not 'return' when you call them. Meaning they're still existing in state and bloating the database.

Make sure you block zero pool accounts from being set and understand when a pool may become zero balance

andrewnguyen22:
This is correct and looks good. Can you do the same for `SetAccount`? Thanks

pablocampogo:
> This is correct and looks good. Can you do the same for `SetAccount`? Thanks

Yes I'll do it

Issue #80, opened 2025-02-14
[FEATURE] Implement Generation and verification of merkle proofs in the sparse merkle tree


rem1niscence:
### Describe the feature you'd like
Implement support for generating and verifying Merkle proofs in a Sparse Merkle Tree (SMT) data structure. This feature should enable both inclusion proofs (to prove the existence of a key-value pair in the tree) and non-inclusion proofs (to prove the absence of a key in the tree). 

### Motivation
I'm crucial to be able to prove the inclusion and non inclusion of data in a sparse data set. By implementing Merkle proof generation and verification for the Sparse Merkle Tree on `StateCommit` store, the blockchain can achieve efficient and secure validation of state transitions

### Possible implementation
(Optional) If you have ideas for how this feature could be implemented, share them here.

#### Required Tags
- Priority: High
- Module: Persistence

PR #100, opened 2025-02-21
Issue #80 - Generate and verify Sparse Merkle tree Proof

rem1niscence:
## Description
Implement the generation of the Sparse Merkle Tree proof and verification of it

## Related Issues

Closes: #80 

## Changes Made
- Fix small issue on wallet package.json across the project
- Add a `proof.proto` file with the `MerkleProof` structure in holder to hold the data in regards of working with merkle proofs
- Implement `GetMerkleProof` to generate a merkle proof for a given key which may or not exist
- Implement `VerifyProof` to verify a proof generatede with `GetMerkleProof` (or in practice any proof for that matter as long as it adheres to the required Proof format)

## Checklist


lib/.proto/proof.proto
```diff
@@ -0,0 +1,48 @@
...
+// For the array of Nodes, the first two nodes represent the initial leaf node siblings
+// required to compute the parent node's hash. The remaining nodes are the intermediate 
+// node siblings necessary to compute the root hash.
+message MerkleProof {
+  // Nodes: Element in the sparse Merkle tree that are required to compute the root hash, augmented with
+  // the compact key used to traverse the tree
+  repeated ProofNode Nodes = 1;
+  // Root: is the cryptographic hash of the root of the Sparse Merkle Tree at the time of 
+  // the proof
+  bytes Root = 8;
```
andrewnguyen22, line 32: I think this can be omitted

rem1niscence: is more like a helper data but if is strictly required that the user may provide the root themselves by an argument then yes it can be removed

lib/.proto/proof.proto
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// ProofNode represents Canopy's implementation a node of the  Sparse Merkle Tree implementation.
+// It is the same as the Node implemented in the store, with the addition of the Key field which
+// stores the compact implmentation of the node's key bit sequences. For more information check the
+// `key` struct in the `store/smt.go` file.
+message ProofNode {
```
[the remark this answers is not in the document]
rem1niscence: done, I had to move the node structure to `lib` in order to avoid a circular dependency

store/smt.go
```diff
@@ -647,13 +814,13 @@ func (x *node) setChildren(leftKey, rightKey []byte) {
 	x.LeftChildKey, x.RightChildKey = leftKey, rightKey
 }
 
-// getOtherChild() returns the sibling for the child key passed
-func (x *node) getOtherChild(childKey []byte) []byte {
+// getOtherChild() returns the sibling for the child key passed and which child it is
+func (x *node) getOtherChild(childKey []byte) ([]byte, byte) {
```
andrewnguyen22, line 818: Should `other child` be an int32?

rem1niscence: if you mean the second argument then it can totally be, I chose `byte` as it only represent one of two possible values (left or right)

store/smt.go
```diff
@@ -419,6 +428,165 @@ func (s *SMT) validateTarget() lib.ErrorI {
...
+// reconstructing the root hash and comparing it against the provided root hash
+// depending on the proof type (membership or non-membership)
+func (s *SMT) VerifyProof(key []byte, value []byte, validateMembership bool, root []byte, proof []*lib.Node) (bool, lib.ErrorI) {
+	// A valid proof is expected to have at least two nodes (the leaf nodes) in order to build the root
+	// and always should include an even number of nodes representing the siblings of the nodes in the tree
+	if len(proof) < 2 && len(proof)%2 != 0 {
+		return false, ErrInvalidMerkleTreeProof()
+	}
+
+	// Generate the inital hash from the leaves of the proof
```
andrewnguyen22, line 496:
```suggestion
	// Generate the initial hash from the leaves of the proof
```

store/smt.go
```diff
@@ -246,15 +258,12 @@ func (s *SMT) traverse() (err lib.ErrorI) {
 		}
 		// assert current key isn't nil
 		if currentKey == nil {
```
andrewnguyen22, line 260: I believe this check should happen prior to 

```go
s.current, err = s.getNode(currentKey)
		if err != nil {
			return
		}
```

store/smt.go
```diff
@@ -246,15 +258,12 @@ func (s *SMT) traverse() (err lib.ErrorI) {
...
-			panic("nil current key")
+			return ErrInvalidMerkleTree()
 		}
 		// load the bytes into the key
 		s.current.Key.fromBytes(currentKey)
```
andrewnguyen22, line 264: It appears as if this call is unnecessary as `s.current, err = s.getNode(currentKey)` does this

store/smt.go
```diff
@@ -419,6 +428,165 @@ func (s *SMT) validateTarget() lib.ErrorI {
...
+	proof := make([]*lib.Node, 0)
+
+	// navigates the tree downward
+	if err := s.traverse(); err != nil {
+		return nil, err
+	}
+
+	// Add current to the list of traversed nodes (actual node in case of proof of membership),
+	// possible location in the case of proof of non-membership
+	s.traversed.Nodes = append(s.traversed.Nodes, s.current.copy())
```
andrewnguyen22, line 450: I'm worried this will panic if `s.current` is nil

andrewnguyen22: Although upon second check, I'm not sure current can be nil because the traversal should return an error.

Is there a case where `s.current` could be nil during a proof of `non-membership`?

rem1niscence: no, because even in non membership on a default tree (root, min and max) on non membership `current` will be equal to either min or max depending on where it would be added

store/smt.go
```diff
@@ -419,6 +428,165 @@ func (s *SMT) validateTarget() lib.ErrorI {
...
+	return proof, nil
+}
+
+// VerifyProof verifies a Sparse Merkle Tree proof for a given value
+// reconstructing the root hash and comparing it against the provided root hash
+// depending on the proof type (membership or non-membership)
+func (s *SMT) VerifyProof(key []byte, value []byte, validateMembership bool, root []byte, proof []*lib.Node) (bool, lib.ErrorI) {
+	// A valid proof is expected to have at least two nodes (the leaf nodes) in order to build the root
+	// and always should include an even number of nodes representing the siblings of the nodes in the tree
+	if len(proof) < 2 && len(proof)%2 != 0 {
```
andrewnguyen22, line 492: Extract out `len` to prevent double call

andrewnguyen22: This should be an `||` I believe

store/smt.go
```diff
@@ -419,6 +428,165 @@ func (s *SMT) validateTarget() lib.ErrorI {
...
+	}
+
+	// Rebuild a similar merkle tree using the proof nodes, to be able to traverse it
+	// again and confirm whether the given key and value are part of the tree, to
+	// confirm the proof-of-non-membership as if the key is not part of the tree
+	memStore, err := NewStoreInMemory(lib.NewDefaultLogger())
+	if err != nil {
+		return false, err
+	}
+	smt := NewSMT(s.Root(), s.keyBitLength, memStore)
```
andrewnguyen22, line 533: I believe this should use `NewDefaultSMT` because s.Root() will clone the current root `value` not `key`

store/smt.go
```diff
@@ -419,6 +428,165 @@ func (s *SMT) validateTarget() lib.ErrorI {
...
+
+	// add the nodes
+	for _, leaf := range proof {
+		// Keys are saved "as-is" to preserve the original values of the tree
+		// when the proof was obtained. Some intermediate node keys may have a length
+		// shorter than the MaxKeyBitLength. This pads such keys to ensure they are
```
andrewnguyen22, line 539: I'm doubtful that this is what we want to do here.

I think we want to write the keys 'as is' without padding

andrewnguyen22: Shouldn't we just use `setNode` for each?

```go
        // set the node in the database
	if err := s.setNode(s.target); err != nil {
		return err
	}
```

store/smt.go
```diff
@@ -419,6 +428,165 @@ func (s *SMT) validateTarget() lib.ErrorI {
...
+		return false, err
+	}
+	// navigates the tree downward
+	if err := smt.traverse(); err != nil {
+		return false, err
+	}
+
+	// Verify whether the key exists in the tree and what kind of proof is being validated
+	// (membership or non-membership).
+	// if the key does not exists in the tree and the proof is for membership or
```
andrewnguyen22, line 562: `does not exist`

store/smt.go
```diff
@@ -419,6 +428,165 @@ func (s *SMT) validateTarget() lib.ErrorI {
...
+	// if the key exists in the tree and the proof is for non-membership, return false.
+	nodeExists := smt.target.Key.equals(smt.gcp)
+	if (!nodeExists && validateMembership) || (nodeExists && !validateMembership) {
+		return false, nil
+	}
+	// if the key does not exists in the tree and the proof is for non-membership, return true.
```
andrewnguyen22, line 568: `does not exist`

store/smt.go
rem1niscence: I could try that, the original issue is that because some intermediate keys were made with a gcp lower than the keyBitLength, generating those keys would panic as the index created by keybitlength would be larger than the gcp length

store/smt.go
```diff
@@ -94,7 +95,7 @@ type node struct {
 	// Key: the structure that is used to interpret node keys (bytes, fromBytes, etc.)
 	Key *key
 	// Node: is the structure persisted on disk under the above key bytes
-	Node
+	lib.Node
```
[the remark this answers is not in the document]
rem1niscence: for deprecating the `node` structure in favour of `lib.Node` I'd need to either add the `key` struct into the `store.proto` and augment the generated struct with the methods that are already implemented on `key` (`fromBytes, bytes`...) or again create the struct in proto but modify the `key` from methods to functions so they don't hold any state but expect external to be passed so they perform their operations, wdyt

or keep it as-is and do not replace `node` for `lib.Node`, I think both solutions above are a bit of a "force" to keep a single structure when the `Key` in `lib.Node` although is generated to use it in merkle proofs, is doing nothing more than storing the data in order to be transmitted somewhere else like `Bitmask`

store/smt.go
andrewnguyen22: You removed the wrong part. Revert

andrewnguyen22: Ensure you fully understand the review comment before making these changes

store/smt.go
rem1niscence: solved

andrewnguyen22 (approved):

Issue #102, opened 2025-02-23
[FEATURE] Re-architect P2P `sends` to NOT use a `sendInterval`


andrewnguyen22:
### Describe the feature you'd like
Re-design the P2P module send service so that it uses channel(s) to trigger sends not polling

```go
func (c *MultiConn) startSendService() {
	defer lib.CatchPanic(c.log)
	send, m := time.NewTicker(sendInterval), limiter.New(0, 0) // INTERVAL IS WASTEFUL
	...
	for {
		// select statement ensures the sequential coordination of the concurrent processes
		select {
		case <-send.C: // fires every 'sendInterval'
			if packet := c.getNextPacket(); packet != nil {
				c.log.Debugf("Send Packet(ID:%s, L:%d, E:%t)", lib.Topic_name[int32(packet.StreamId)], len(packet.Bytes), packet.Eof)
				err = c.sendWireBytes(packet, m)
			}
```

Feature checklist
- [ ] Solution does not use interval checks to 'see' if a packet is available to send
NOTE: currently our P2P send service does all of the below, so this must be maintained with the new architecture
- [ ] Allows `pings` and `pongs` to be sent in between packets (even for the same message)
- [ ] Allows `stream switching` (when sending a message the multiplexed stream may change)
- [ ] Enables `stream priority` (when sending a message, certain streams are prioritized over others)

### Motivation
- intervals waste CPU cycles
- intervals make 'debug logging' difficult
- intervals may 'backlog'

#### Required Tags
- Priority: Medium
- Module: P2P

PR #111, opened 2025-03-07
feat: change message send to use a controlled concurrency queue instead of interval

pablocampogo:
## Description
Changes send interval to use a general `sendQueue` that directly sends a packet when it is queued, controling the concurrency with a semaphore.

## Related Issues
Closes: #102

## Changes Made
- Add `sendQueue` channel to `MultiConn` struct
- Add new constants for max size of `sendQueue` and concurrency
- Add semaphore to control concurrency in the queue
- Deleted send interval logic

## Checklist
- [] I have updated documentation (if applicable).


andrewnguyen22 (changes requested):
This is a clever implementation that has promise. A few things:

1) The implementation has a logical flaw where it is `triggering` a send each `message` but completing the trigger once a `packet` of that message is sent. There may be 1 or more packets per message. This will result in many packets backing up in the channel without corresponding triggers. Please be sure to look carefully at the send logic in order to fully understand the flow `hasStuffToSend()` and `nextPacket()` are good starting places

2) The naming `sendQueue` is misleading for the channel. It's more like a `notifySend`

```go
select {
    case c.sendQueue <- struct{}{}:
        return true
    case <-time.After(queueSendTimeout):
        return false
    }
```

3) Unhandled error on semaphore: `sem.Acquire(context.Background(), 1)`

ezeike:
Some comments:

1. Using goroutines and semaphores in this way results in a queue consisting of goroutines waiting for the semaphore to be released. IMO we should stick to more standard ways of queuing items.
2. Having these goroutines running outside of the control of the for-select loop in `startSendService` results in the possibility of `sendWireBytes` being called concurrently. Go's `net.Conn.Write` is not safe for concurrent access.

pablocampogo:
> Some comments:
> 
> 1. Using goroutines and semaphores in this way results in a queue consisting of goroutines waiting for the semaphore to be released. IMO we should stick to more standard ways of queuing items that have visibility and limits to their size.
> 2. Having these goroutines running outside of the control of the for-select loop in `startSendService` results in the possibility of `sendWireBytes` being called concurrently. Go's `net.Conn.Write` is not safe for concurrent access.
> 
> One idea:
> 
> `sendQueue` channel is modified to only have a buffer size of 1 and is of type Packet. This channel is kept full by a new goroutine. This way the current for-select loop still multiplexes properly between packets/pings/pongs and still allows for stream switching & stream priority.
> 
> The `Send` method would then be modified to send *Envelope to this new goroutine.

I think the semaphore approach is more readable, and I've used similar approaches before for really heavy concurrent processes (indexing) and it has worked just fine.

On the other hand yes you're right `net.Conn.Write` is not concurrent safe, I'll add a new writer channel with a buffer of 1 to manage that part.

p2p/conn.go
```diff
@@ -116,30 +131,114 @@ func (c *MultiConn) Send(topic lib.Topic, msg *Envelope) (ok bool) {
...
+	default:
+		c.log.Debugf("Send Packet(ID:%s, L:%d, MID:%s)", lib.Topic_name[int32(p.StreamId)], len(p.Bytes), p.MessageId)
+		err := c.sendWireBytes(p, m)
+		if err != nil {
+			c.sendErrChan <- err
+		}
+	}
+}
+
+func (c *MultiConn) startWriteChannel() {
```
andrewnguyen22, line 205: Add function comment

p2p/conn.go
```diff
@@ -116,30 +131,114 @@ func (c *MultiConn) Send(topic lib.Topic, msg *Envelope) (ok bool) {
...
+	for data := range c.writeCh {
+		_, err := c.conn.Write(data)
+		if err != nil {
+			c.log.Error(ErrFailedWrite(err).Error())
+			c.sendErrChan <- ErrFailedWrite(err)
+		}
+	}
+}
+
+func (c *MultiConn) queueWrite(b []byte) bool {
```
andrewnguyen22, line 216: add function comment

andrewnguyen22 (changes requested):
Our main principle of our design is to make everything as simple as possible. With this framing - we'll likely need to make some adjustments to this implementation. 

The good about this PR:
  A) It works!
  B) It removed the need for a *send ticker
  C) Demonstrates a good understanding of the module

The main issue with this PR is:
  A) It introduces overhead ex. `packetHeader` went from 47 bytes to 103
  B) It complicates the code ex. *introduces several new channels, several fields, and a new write service*
  C) Needs more comments

See specifics below

1) Why should we use multi-threading to send? 

Our goal isn't to send as fast as possible - in-fact, we want to rate limit ourselves to not 'over send' to our peers.

I believe this introduces the need for:
  A) A semaphore
  B) A UUID to re-assemble the packets on the other side
  C) An additional service to do writes

These introductions are all undesirable because:
  A) It's difficult to follow the flow
  B) It requires more resources
  C) The packets have more overhead

Let me know if there are other reasons why this was done and we can discuss further

2) Why do we need to 'notify' of a packet queued?

```go
        ok = stream.queueSend(packet)
        ok = c.notifySend()
```

I believe queuing the packet itself is a notification and can be leveraged.

This is more desirable because:
  A) It's simpler
  B) It doesn't require an additional channel for notifications

3) Why are packets needing to be 'indexed' for order?

This doesn't seem to be necessary because:
  A) We're using `TCP` not `UDP` so we already have a reliable packet transfer and ordering method
  B) There's no 'recovery' mechanism implemented if a packet were to be lost

This addition is undesirable because:
  A) It complicates the msgAssembler: `msgAssembler map[string]map[int][]byte`
  B) It complicates the `handlePacket` function
  C) It introduces even more overhead to the packets

pablocampogo:
> Our main principle of our design is to make everything as simple as possible. With this framing - we'll likely need to make some adjustments to this implementation.
> 
> The good about this PR: A) It works! B) It removed the need for a *send ticker C) Demonstrates a good understanding of the module
> 
> The main issue with this PR is: A) It introduces overhead ex. `packetHeader` went from 47 bytes to 103 B) It complicates the code ex. _introduces several new channels, several fields, and a new write service_ C) Needs more comments
> 
> See specifics below
> 
> 1. Why should we use multi-threading to send?
> 
> Our goal isn't to send as fast as possible - in-fact, we want to rate limit ourselves to not 'over send' to our peers.
> 
> I believe this introduces the need for: A) A semaphore B) A UUID to re-assemble the packets on the other side C) An additional service to do writes
> 
> These introductions are all undesirable because: A) It's difficult to follow the flow B) It requires more resources C) The packets have more overhead
> 
> Let me know if there are other reasons why this was done and we can discuss further
> 
> 2. Why do we need to 'notify' of a packet queued?
> 
> ```go
>         ok = stream.queueSend(packet)
>         ok = c.notifySend()
> ```
> 
> I believe queuing the packet itself is a notification and can be leveraged.
> 
> This is more desirable because: A) It's simpler B) It doesn't require an additional channel for notifications
> 
> 3. Why are packets needing to be 'indexed' for order?
> 
> This doesn't seem to be necessary because: A) We're using `TCP` not `UDP` so we already have a reliable packet transfer and ordering method B) There's no 'recovery' mechanism implemented if a packet were to be lost
> 
> This addition is undesirable because: A) It complicates the msgAssembler: `msgAssembler map[string]map[int][]byte` B) It complicates the `handlePacket` function C) It introduces even more overhead to the packets

1.. If we want to keep sending one by one we can, I can do a commit to do the change back, this would simplify the implementation.
2. I'm pretty sure at this point I can find a way to not need that extra channel, ie maybe iterating over the topics instead of just looking to one single channel.
3. If we want to go back to one by one, this is not needed anymore. Receive logic can go back to what it was before and the Packet struct can go back too.

If we want to keep my approach, one change would be to set the int to bytes in a way it takes less bytes and do the change to not need that extra channel.

Should I go back to the linear approach?

andrewnguyen22:
> > Our main principle of our design is to make everything as simple as possible. With this framing - we'll likely need to make some adjustments to this implementation.
> > The good about this PR: A) It works! B) It removed the need for a *send ticker C) Demonstrates a good understanding of the module
> > The main issue with this PR is: A) It introduces overhead ex. `packetHeader` went from 47 bytes to 103 B) It complicates the code ex. _introduces several new channels, several fields, and a new write service_ C) Needs more comments
> > See specifics below
> > 
> > 1. Why should we use multi-threading to send?
> > 
> > Our goal isn't to send as fast as possible - in-fact, we want to rate limit ourselves to not 'over send' to our peers.
> > I believe this introduces the need for: A) A semaphore B) A UUID to re-assemble the packets on the other side C) An additional service to do writes
> > These introductions are all undesirable because: A) It's difficult to follow the flow B) It requires more resources C) The packets have more overhead
> > Let me know if there are other reasons why this was done and we can discuss further
> > 
> > 2. Why do we need to 'notify' of a packet queued?
> > 
> > ```go
> >         ok = stream.queueSend(packet)
> >         ok = c.notifySend()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I believe queuing the packet itself is a notification and can be leveraged.
> > This is more desirable because: A) It's simpler B) It doesn't require an additional channel for notifications
> > 
> > 3. Why are packets needing to be 'indexed' for order?
> > 
> > This doesn't seem to be necessary because: A) We're using `TCP` not `UDP` so we already have a reliable packet transfer and ordering method B) There's no 'recovery' mechanism implemented if a packet were to be lost
> > This addition is undesirable because: A) It complicates the msgAssembler: `msgAssembler map[string]map[int][]byte` B) It complicates the `handlePacket` function C) It introduces even more overhead to the packets
> 
> 1.. If we want to keep sending one by one we can, I can do a commit to do the change back, this would simplify the implementation. 2. I'm pretty sure at this point I can find a way to not need that extra channel, ie maybe iterating over the topics instead of just looking to one single channel. 3. If we want to go back to one by one, this is not needed anymore. Receive logic can go back to what it was before and the Packet struct can go back too.
> 
> If we want to keep my approach, one change would be to set the int to bytes in a way it takes less bytes and do the change to not need that extra channel.
> 
> Should I go back to the linear approach?

Let's go back to linear @pablocampogo 

Thanks

pablocampogo:
@andrewnguyen22 the only way of doing this without needing an extra notify channels is using reflect, otherwise I would need to do an ugly select case that hard coded includes each topic channel.

We have two options right now:
1. Do it with the reflect package, as it is currently in this PR.
2. Change to what previously discussed of just having one channel for all the topics and prioritize through a map at a new "level".

I lean more on option no.1 but tell me what you think.

andrewnguyen22:
> @andrewnguyen22 the only way of doing this without needing an extra notify channels is using reflect, otherwise I would need to do an ugly select case that hard coded includes each topic channel.
> 
> We have two options right now:
> 
> 1. Do it with the reflect package, as it is currently in this PR.
> 2. Change to what previously discussed of just having one channel for all the topics and prioritize through a map at a new "level".
> 
> I lean more on option no.1 but tell me what you think.

Not sure what this looks like with reflect? Can't we switch on the topic?

Also what are your thoughts on #122

andrewnguyen22:
Ah ok - it's in the code. One moment

pablocampogo:
> > @andrewnguyen22 the only way of doing this without needing an extra notify channels is using reflect, otherwise I would need to do an ugly select case that hard coded includes each topic channel.
> > We have two options right now:
> > 
> > 1. Do it with the reflect package, as it is currently in this PR.
> > 2. Change to what previously discussed of just having one channel for all the topics and prioritize through a map at a new "level".
> > 
> > I lean more on option no.1 but tell me what you think.
> 
> Not sure what this looks like with reflect? Can't we switch on the topic?
> 
> Also what are your thoughts on #122

Although 122's implementation seems it will work, my implementation at the didn't add any new channels, more similar to the original implementation.

andrewnguyen22 (approved):

andrewnguyen22:
Rolling back due to alphanet halt

Issue #97, opened 2025-02-21
[FEATURE] Clean up the `server.go` file for the rpc


andrewnguyen22:
### Describe the feature you'd like
Currently the `cmd/rpc/server.go` file for the rpc lacks the hygiene standards of the rest of the application.

- Add a comment (nearly) per line of code
- Look for opportunities to separate the RPC payload structures
- Fix the `memory leak` TODO by re-architecting the code

```go
// TODO likely a memory leak here from un-discarded stores
func setupStateMachine(height uint64, w http.ResponseWriter) (*fsm.StateMachine, bool) {
	state, err := app.FSM.TimeMachine(height)
	if err != nil {
		write(w, ErrTimeMachine(err), http.StatusInternalServerError)
		return nil, false
	}
	return state, true
}
```

### Motivation
Currently the `cmd/rpc/server.go` file for the rpc lacks the hygiene standards of the rest of the application. As it's the interface file, many client-side application builders will look here first to understand the API. Currently it is not a good representation of the cleanliness of the code.

### Additional context
- Affects wallet and block explorer so be sure to fix breaking changes
- **After completion of this issue** the RPC should be finalized and ready for specification.

#### Required Tags
- Priority: High
- Module: RPC

andrewnguyen22:
@pablocampogo must finish #34 before this issue may be completed

ezeike:
In addition to the improvements identified in the issue description:

- Code Comments & Code Quality Practices
- Remote Procedure Call (RPC) Payload Structures
- Resource Management Issues in TimeMachine

Additional Observations Noted or Discussed:

- All RPC server functionality is contained within a single monolithic source file.
- Route definitions are currently stored as string literals, which affects the clarity and re-usability of route configurations.
- Reliance on global variables for critical data elements such as database connections, controllers, and configuration settings.
- RPC server instantiates an RPC client to query itself resulting in tight coupling between the server and client types.
- Database access requires consistent usage of `Discard` with `store.NewStoreWithDB` as well as the `TimeMachine`. While `Discard` is being correctly called after `Store` access, applying the same solution for `TimeMachine` may be beneficial here.
- The CLI tool utilizes `StringToCommittees`, but there exists a duplicate and similarly named function `stringToCommittees` within this package, potentially leading to confusion or future mistakes.

ezeike:
**Proposed PR Order:**

---

**Phase One:**

- Implement changes that do not alter logic or program flow.
- Refactor route configurations for improved clarity.
[...]

PR #114, opened 2025-03-07
server.go global variable refactor + comment updates

ezeike:
A PR implementing most of the changes for issue-https://github.com/canopy-network/canopy/issues/97 Phase One.

    Implement changes that do not alter logic or program flow.
    Refactor route configurations for improved clarity.
    Refactor code to eliminate global variables.
    Separate query and admin request handlers into distinct files, specifically admin_handlers.go and query_handlers.go.
    Add comments to affected sections of the code base.
    Refactor the startup logic to enhance clarity and maintainability.

The newly introduced Server type will encapsulate what were previously global
variables. Consequently, many functions have been converted into methods
associated with the Server type due to their reliance on these global variables.


cmd/rpc/admin_handlers.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+
+	// Retrieve the private key for the signer from the keystore using the provided password.
+	privateKey, err := keystore.GetKey(signer, ptr.Password)
+	if err != nil {
+		write(w, err, http.StatusBadRequest)
+		return
+	}
+	// Set the public key in transaction request for reference.
+	ptr.PubKey = privateKey.PublicKey().String()
```
andrewnguyen22, line 454: If you port over this from main

```go
operatorPrivateKey, err := keystore.GetKey(ptr.Address, ptr.Password)
	if err != nil {
		write(w, err, http.StatusBadRequest)
		return
	}
	ptr.PubKey = operatorPrivateKey.PublicKey().String()
	p, err := callback(signerPrivateKey, ptr)
	if err != nil {
		write(w, err, http.StatusBadRequest)
		return
	}
```

Then your conflict is likely non-existent in the other file

PR #123, opened 2025-03-11
txRequest type updates for client.go

ezeike:
## Description
Phase 3 of Issue #97.

Client side refactor of RPC transaction types.

## Related Issues
Closes: #97

## Changes Made
- Monolithic txRequest type split into Tx-specific request types, mirroring how other non-Tx types are implemented.
- Removed single-field struct embeds such as passwordRequest.
- New `fromRequest` and `signerRequest` type created to simplify some handling
- New functions `getFrom` and `getSigner` to simplify some handling, obsoleting old `setFrom` and `setSigner`.
- Submit field moved out of Tx types.

## Checklist
- [] I have updated documentation (if applicable).
- [] I have included tests for the changes (if applicable).

### DIFF
--- a/cmd/rpc/client.go
+++ b/cmd/rpc/client.go
@@ -4,9 +4,10 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/canopy-network/canopy/fsm"
 	"io"
 	"net/http"
+
+	"github.com/canopy-network/canopy/fsm"
 
 	"github.com/canopy-network/canopy/controller"
 	"github.com/canopy-network/canopy/lib"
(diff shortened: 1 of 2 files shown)


andrewnguyen22 (commented):
First pass - this all looks good. Will test the transactions later

rem1niscence (commented):
This PR looks great! Just one suggestion for further improvement: since this PR is already working with updates to the txRequest type, I noticed that most (if not all) of the modified functions related to custom tx structs return a `transactionRequest`, which in turn includes three variables. Would it make sense to consolidate `hash *string` and `tx json.RawMessage` into a `transactionRequest` struct as well?

This change could help streamline the code further, especially since many of these functions already handle a significant number of arguments. By embedding these variables into the `transactionRequest` struct, we could simplify the function signatures and potentially improve readability just a bit more

ezeike:
> This PR looks great! Just one suggestion for further improvement: since this PR is already working with updates to the txRequest type, I noticed that most (if not all) of the modified functions related to custom tx structs return a `transactionRequest`, which in turn includes three variables. Would it make sense to consolidate `hash *string` and `tx json.RawMessage` into a `transactionRequest` struct as well?
> 
> This change could help streamline the code further, especially since many of these functions already handle a significant number of arguments. By embedding these variables into the `transactionRequest` struct, we could simplify the function signatures and potentially improve readability just a bit more

`transactionRequest` is actually a function that returns two values plus an error. As it is now creating a type for these two values would make it harder to read IMO.

rem1niscence (approved):

Issue #116, opened 2025-03-08
[BUG] `Max` doesn't account for the transaction fee


andrewnguyen22:
### Describe the bug
Using the max button in the wallet fails because it doesn't account for the transaction fee

<img width="815" alt="Image" src="https://github.com/user-attachments/assets/b18c8dee-2e19-409d-bce9-c07734f780d1" />

### Steps to reproduce
1. Start a send transaction
2. Click on `Max`
3. Attempt to send

### Expected behavior
The max button should subtract the 'fee amount'

### Environment
- **Browser** Chrome

#### Required Tags
- Priority: Medium
- Module: Wallet

andrewnguyen22:
NOTE: This issue should be tackled with #73 as 'auto-filling' the transaction fee will allow us to determine the 'true max'

PR #125, opened 2025-03-12
Issue-116: default fee on txs and substract from max

rem1niscence:
## Description

Sets the default fee as a default value in the tx forms and substracts that amount from the available max balance

## Related Issues

Closes: #116
Closes: #73

## Checklist

## Additional Notes

Changes are actually less than what's shown, most of them are due to linter changes


cmd/rpc/web/wallet/components/util.js
```diff
@@ -663,3 +663,37 @@ export async function retryWithDelay(fn, onFailure, retries = 8, delayMs = 1000,
...
+export function getActionFee(action, params) {
+  switch (action) {
+      case "send":
+          return params.sendFee;
+      case "stake":
+          return params.stakeFee;
+      case "create_order":
+          return params.createOrderFee;
+      case "lock_order":
+          return params.lockOrderFee;
```
andrewnguyen22, line 677: There is no lock order fee parameter - was this tested?

andrewnguyen22 (changes requested):
Your front end PRs change a lot of the formatting and it's difficult to review. Can you make adjustments so this doesn't continue to happen? See diff

cmd/rpc/web/wallet/components/util.js
rem1niscence: I did test them but still interesting, I think that I read about it on the params shown in the explorer (which I got the info from), seems like I read wrong as I was adding them on my side, fixing

rem1niscence:
> Can you make adjustments so this doesn't continue to happen? 
Issue happens because the multiple contributors have different linter configurations instead of following the prettier config which is set on the project, one way of enforcing this is to configure precommit hooks so they run prettier after each commit to these files, I can do that but let me know if you would like that solution as we would need to update the README to ask developers to install `pre-commit`

rem1niscence:
The PR has been updated by running Prettier again. Moving forward, Prettier will be required to run before any new PRs are submitted, and this will be enforced for all upcoming PRs.

Issue #105, opened 2025-02-24
[FEATURE] RPC endpoint and dynamic Explorer display for minted CNPY per block


andrewnguyen22:
### Describe the feature you'd like

Feature Checklist
- [ ] Create an RPC endpoint for 'economic parameters'
- - Mint per block
- - Mint per committee
- - DAO cut
- - Block proposer cut
- - Delegate cut

- [ ] Dynamically display the mint per block on the block explorer

<img width="315" alt="Image" src="https://github.com/user-attachments/assets/f86b08ff-ad70-4fd5-8b98-8aa85c2de738" />
 

### Motivation
- Currently the display for minted CNPY is fixed, it should be dynamic
- This feature will certainly answer future questions regarding the economics

#### Required Tags
- Priority: Medium
- Module: RPC, Explorer

PR #129, opened 2025-03-17
Economic Parameters endpoint

ezeike:
## Description

Creates an economic parameters endpoint that provides the following:

- Mint per block
- Mint per committee
- DAO cut
- Block proposer cut
- Delegate cut

Explorer updated to dynamically update the minted CNPY per block.

## Related Issues

Closes: #105 

## Checklist

### DIFF
--- a/cmd/rpc/types.go
+++ b/cmd/rpc/types.go
@@ -14,6 +14,10 @@ type heightRequest struct {
 	Height uint64 `json:"height"`
 }
 
+type chainRequest struct {
+	ChainId uint64 `json:"chainId"`
+}
+
 type orderRequest struct {
 	ChainId uint64 `json:"chainId"`
 	OrderId uint64 `json:"orderId"`
@@ -134,6 +138,14 @@ type addressRequest struct {
 
 type committeesRequest struct {
 	Committees string
+}
+
+type economicParameterResponse struct {
+	MintPerBlock     uint64 `json:"MintPerBlock"`
+	MintPerCommittee uint64 `json:"MintPerCommittee"`
+	DAOCut           uint64 `json:"DAOCut"`
+	ProposerCut      uint64 `json:"ProposerCut"`
+	DelegateCut      uint64 `json:"DelegateCut"`
 }
 
 // =====================================================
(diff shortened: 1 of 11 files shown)


andrewnguyen22:
@ezeike is this still a draft?

fsm/committee.go
```diff
@@ -34,20 +34,40 @@ func (s *StateMachine) FundCommitteeRewardPools() lib.ErrorI {
...
+	mintAmountPerCommittee = mintAmountAfterDAOCut / subsidizedCount
+
+	return daoCut, totalMintAmount, mintAmountPerCommittee, nil
+}
+
+// FundCommitteeRewardPools() mints newly created tokens to protocol subsidized committees
+func (s *StateMachine) FundCommitteeRewardPools() lib.ErrorI {
+	daoCut, _, mintAmountPerCommittee, err := s.GetBlockMintStats(s.Config.ChainId)
+	if err != nil {
+		if err.Error() == "Nothing to do" {
```
rem1niscence, line 57: same here, when changing to a pre-existing error from `fsm/errors.go` please do compare it against the error using `errors.Is`

cmd/rpc/types.go
```diff
@@ -14,6 +14,10 @@ type heightRequest struct {
 	Height uint64 `json:"height"`
 }
 
+type chainRequest struct {
+	ChainId uint64 `json:"chainId"`
```
rem1niscence, line 18: is this meant to always query the chain that the node is running or also the root chain? because if is only the chain that the node is running then it may be unnecesary to pass the `chainID`

rem1niscence (commented):
Great PR overall!!, just a few comments about conventions and a possible removal of the chainID param

andrewnguyen22:
@ezeike can we get this in soon?

cmd/rpc/types.go
ezeike: This wasn't specified in the issue but through some questions with Andrew it looked like we should be able to query both.

rem1niscence (approved):

PR #132, opened 2025-03-26
feat: change P2P handle send message logic

pablocampogo:
PR just for visibility of test deploy


pablocampogo:
> what's the status of the mutex change? did it proved to solve the issue or pinpoint at what the possible issue is?

I commented the mutex for now because they were being too slow on prod, just leaving it running with the extra logs until the issue happens again.

PR #142, opened 2025-04-10
hotfix: change recover panics to not use helpers and be called directly in func

pablocampogo:
Change panics recover to be called directly in func and not use helper function


rem1niscence:
please solve the conflict

rem1niscence:
also I see the changes here are also applied in #142, close this PR if so to avoid duplication or more conflicts

andrewnguyen22 (approved):

PR #143, opened 2025-04-13
Partition DB Design

andrewnguyen22:
## Historical State Partitioning (BadgerDB, Managed Mode)

Using BadgerDB in **managed mode**, we treat block heights as timestamps. This allows precise versioning of key/value pairs at each block.

### Prefixes

- `/State/` – latest state
- `/<PartitionPrefix>/` – historical snapshots (every N blocks)
- `N = 1000`
- `PartitionPrefix = (Height / N) * N` (e.g., 4787 → 4000)

---

### Step 1: Double Write on Set

On each new write at `LatestHeight`:

- `/State/<Key>` → `Value` @ `LatestHeight`
- `/<PartitionPrefix>/<Key>` → `Value` @ `LatestHeight`

---

### Step 2: Flatten Snapshot Boundary

If `LatestHeight % N == 1`:

- `LastHeight = LatestHeight - 1`
- For each key in `/State/` at version `LastHeight`:
  - Copy to `/<PartitionPrefix>/<Key>` at `LastHeight`
  - Delete all older `/State/<Key>` versions

This ensures snapshots are complete for each partition.

---

### Step 3: Historical Queries

To read key at `targetBlock`:

```go
txn := db.NewTransactionAt(targetBlock)

if LatestHeight - targetBlock < N:
    prefix = /State/
else:
    PartitionPrefix = (targetBlock / N) * N
    prefix = /<PartitionPrefix>/
    
txn.Get(prefix + key)
```

---

### Benefits

- Keeps `/State/` fast to iterate (few versions)
- Historical data is chunked → faster seeks & smaller iterator windows
- No reverse seeks needed due to managed timestamps

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -231,14 +231,14 @@ func (s *Store) Partition() {
 	if err = func() lib.ErrorI {
 		// calculate the partition height
 		snapshotHeight := partitionHeight(sc.Version())
+		// generate the historical partition prefix
+		partitionPrefix := historicalPrefix(snapshotHeight)
 		// create a reader at the partition height
 		reader := sc.db.NewTransactionAt(snapshotHeight, false)
 		defer reader.Discard()
 		// create a managed batch to do the 'writing'
 		writer := sc.db.NewWriteBatchAt(snapshotHeight)
 		defer writer.Cancel()
-		// generate the historical partition prefix
-		partitionPrefix := historicalPrefix(snapshotHeight)
 		// get the latest store in a transaction wrapper
 		lss := NewTxnWrapper(reader, sc.log, []byte(latestStatePrefix))
 		// create an iterator that traverses the entire latest state store at the partition height
--- a/store/store.go
+++ b/store/store.go
@@ -13,15 +13,15 @@ import (
 )
 
 const (
-	stateStorePrefix      = "s/"          // prefix designated for the LatestStateStore where the most recent blobs of state data are held
-	historicStatePrefix   = "h/"          // prefix designated for the HistoricalStateStore where the historical blobs of state data are held
-	stateCommitmentPrefix = "c/"          // prefix designated for the StateCommitmentStore (immutable, tree DB) built of hashes of state store data
-	indexerPrefix         = "i/"          // prefix designated for indexer (transactions, blocks, and quorum certificates)
-	stateCommitIDPrefix   = "x/"          // prefix designated for the commit ID (height and state merkle root)
-	lastCommitIDPrefix    = "a/"          // prefix designated for the latest commit ID for easy access (latest height and latest state merkle root)
-	partitionExistsKey    = "e/"          // to check if partition exists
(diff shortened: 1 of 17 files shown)


store/store.go
```diff
@@ -168,23 +186,118 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
...
+			if e := writer.SetEntryAt((&badger.Entry{Key: append([]byte(stateStorePrefix), k...), Value: v}).WithDiscard(), partHeight); e != nil {
+				return ErrSetBatch(e)
+			}
+		}
+		// commit the batch
+		if e := writer.Flush(); e != nil {
+			return ErrFlushBatch(e)
+		}
+		// trigger garbage collector to prune keys
+		if e := s.db.RunValueLogGC(.25); e != nil {
```
rem1niscence, line 256: nitpick: `RunValueLogGC` usage should be protected with as it states on its description that one of them should run at a time, although is not useful now, adding this would be useful for when adding asynchronous partitioning during syncing or at least a comment for it

store/wrapper_txn.go
```diff
@@ -4,6 +4,16 @@ import (
 	"bytes"
 	"github.com/canopy-network/canopy/lib"
 	"github.com/dgraph-io/badger/v4"
+	"reflect"
+	"unsafe"
+)
+
+const (
+	badgerMetaFieldName               = "meta"    // badgerDB Entry 'meta' field name
+	badgerDiscardEarlierVersions byte = 1 << 2    // badgerDB 'discard earlier versions' flag
```
rem1niscence, line 13: although this codebase is not a guide on how to implement managed badgerdb, linking the [source](https://github.com/hypermodeinc/badger/blob/main/value.go) of these bits will help future readers/maintainers of the codebase as this is now a simple yet integral piece of how the store works

store/wrapper_txn.go
```diff
@@ -4,6 +4,16 @@ import (
...
+	badgerDeleteBit              byte = 1 << 0    // badgerDB 'tombstoned' flag
+	badgerNoDiscardBit           byte = 1 << 3    // badgerDB 'never discard'  bit
+	badgerGCRatio                     = .00000001 // the ratio when badgerDB will run the garbage collector
```
rem1niscence, line 16: any particular reason of setting this to the lowest possible value? my only concern is that, as the documentation implies, it would increase the activity of the underlying LSM tree thus in the worst case affecting the block commit times on validators which may be setup on slower instances

store/wrapper_txn_test.go
```diff
@@ -26,6 +28,52 @@ func TestGetSetDelete(t *testing.T) {
 	require.Contains(t, er.Error(), badger.ErrKeyNotFound.Error())
 }
 
+func TestSetDeleteNonPruneable(t *testing.T) {
```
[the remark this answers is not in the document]
andrewnguyen22: TODO with integration testing

PR #146, opened 2025-04-15
feat: add back ping pong

pablocampogo:
Add back old ping pong implementation using new `sendProtoMsg` func

Dependency: https://github.com/canopy-network/canopy/pull/144 for unit test fix

### DIFF
--- a/p2p/conn.go
+++ b/p2p/conn.go
@@ -16,6 +16,8 @@ import (
 )
 
 const (
+	pingInterval           = 30 * time.Second        // how often a ping is to be sent
+	pongTimeoutDuration    = 20 * time.Second        // how long the sender of a ping waits for a pong before throwing an error
 	maxDataChunkSize       = 1024 - packetHeaderSize // maximum size of the chunk of bytes in a packet
 	maxPacketSize          = 1024                    // maximum size of the full packet
 	packetHeaderSize       = 47                      // the overhead of the protobuf packet header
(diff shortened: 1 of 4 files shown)


p2p/conn.go
```diff
@@ -143,21 +149,64 @@ func (c *MultiConn) Send(topic lib.Topic, msg *Envelope) (ok bool) {
...
 		case packet = <-c.streams[lib.Topic_BLOCK].sendQueue:
+			c.sendPacket(packet, m)
 		case packet = <-c.streams[lib.Topic_BLOCK_REQUEST].sendQueue:
+			c.sendPacket(packet, m)
 		case packet = <-c.streams[lib.Topic_TX].sendQueue:
+			c.sendPacket(packet, m)
 		case packet = <-c.streams[lib.Topic_PEERS_RESPONSE].sendQueue:
+			c.sendPacket(packet, m)
 		case packet = <-c.streams[lib.Topic_PEERS_REQUEST].sendQueue:
+			c.sendPacket(packet, m)
```
rem1niscence, line 170: although this whole logic makes sense to be here as is dealing with sending (and receiving on `startReceiveService`, I think this feels odd that all the types of messages follow the same pattern and ping/pong breaks it completely, how about doing a `startPingPongService` and move the logic for both send and receive there, could call two goroutines there to handle each select and have the timer logic outside as is happening here, do let me know what you think

pablocampogo: I like more for them to be on the already existing send and receive channels so we don't add more channels.

andrewnguyen22: I agree with @pablocampogo I think this is fine

rem1niscence (approved):

Issue #131, opened 2025-03-22
[FEATURE] Syncing Optimization


andrewnguyen22:
### Describe the feature you'd like
- [ ] Don't send the to syncing peers in `gossipBlock()`
- [ ] Improve syncing speeds

### Motivation
Syncing the blockchain is currently quite slow:
- Requests 1 block at a time from 1 peer
- Is interrupted periodically with the *latest* height being sent (gossiped) because the sender doesn't realize the peer is syncing and isn't interested in that height yet.

### Additional context
`/controller/consensus.go`

```go
// Sync() downloads the blockchain from peers until 'synced' to the latest 'height'
// 1) Get the height and begin block params from the state_machine
// 2) Get peer max_height from P2P
// 3) Ask peers for a block at a time
// 4) CheckBasic each block by checking if cert was signed by +2/3 maj (COMMIT message)
// 5) Commit block against the state machine, if error fatal
// 5) Do this until reach the max-peer-height
// 6) Stay on top by listening to incoming cert messages
func (c *Controller) Sync() {
```

### Possible implementation
- Queue many blocks and process them sequentially?
- Request 100 blocks at a time from a single peer and process them all before making the next request?

#### Required Tags
- Priority: Medium
- Module: Consensus

PR #149, opened 2025-04-17
Issue #131 - Node Sync Optimization

ezeike:
# Description

This PR updates the node syncing functionality in `controller/consensus.go` to optimize synchronization and handle unexpected blocks.

# Changes Made

- Implements a queue to track both block requests sent and block request responses awaiting processing by the FSM.
- A new type `blockSyncRequest` was created to track these requests in the queue.
- Block requests are distributed randomly to all known peers in the PeerSet.
- Uses the SimpleLimiter to limit requests per node to avoid rate limiting and reputation slashing.
- A `rateScaleFactor` setting to allow node runners to increase or decrease their block request rates
- Gracefully handles (ignores and prints warning) when the latest gossiped block is received.

# Closes
Issue: #131

### DIFF
--- a/controller/consensus.go
+++ b/controller/consensus.go
@@ -60,11 +60,12 @@ func getRandomAllowedPeer(peers []string, limiter *lib.SimpleLimiter) string {
 // Sync() downloads the blockchain from peers until 'synced' to the latest 'height'
 // 1) Get the height and begin block params from the state_machine
 // 2) Get peer max_height from P2P
-// 3) Ask peers for a block at a time
-// 4) CheckBasic each block by checking if cert was signed by +2/3 maj (COMMIT message)
-// 5) Commit block against the state machine, if error fatal
-// 5) Do this until reach the max-peer-height
-// 6) Stay on top by listening to incoming cert messages
+// 3) Fill a queue of blocks by requesting them from randomly chosen peers
+// 4) When available, process next block from queue
+// 5) CheckBasic each block by checking if cert was signed by +2/3 maj (COMMIT message)
+// 6) Commit block against the state machine, if error fatal
+// 7) Do this until reach the max-peer-height
+// 8) Stay on top by listening to incoming cert messages
 func (c *Controller) Sync() {
 	// log the initialization of the syncing process
 	c.log.Infof("Sync started 🔄 for committee %d", c.Config.ChainId)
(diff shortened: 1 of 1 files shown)


controller/consensus.go
```diff
@@ -35,81 +78,212 @@ func (c *Controller) Sync() {
...
+
+		// Add new request to queue
+		queue[height] = blockSyncRequest{
+			height:        height,
+			peerPublicKey: peerPublicKey,
+		}
+	}
+}
+
+// applyTimeouts removes reuqests from the queue that have timed out
```
rem1niscence, line 241:
```suggestion
// applyTimeouts removes requests from the queue that have timed out
```

controller/consensus.go
```diff
@@ -35,81 +78,212 @@ func (c *Controller) Sync() {
...
+func (c *Controller) applyTimeouts(queue map[uint64]blockSyncRequest) []blockSyncRequest {
```
rem1niscence, line 242: this behavior is not strictly related to the controller itself and I see you perform more operations over the `queue`, why don't convert the `queue` map into its own type like `type queue map[uint64]blockSyncRequest{}` and more this method and queuing/dequeueing to methods of that type, that way you can encapsulate all the behavior of this queue there

rem1niscence (changes requested):

controller/consensus.go
```diff
@@ -35,81 +78,212 @@ func (c *Controller) Sync() {
...
+//   - Block height has advanced and there's room at the end of the queue
+func (c *Controller) sendBlockRequests(start, stop uint64, queue map[uint64]blockSyncRequest, limiter *lib.SimpleLimiter, peers []string) {
+	// Send requests to populate the queue
+	for height := start; height < stop; height++ {
+		// Reached the specified stop height
+		if height >= stop {
+			break
+		}
+		// A block request has already been sent for this height
+		if _, ok := queue[height]; ok {
```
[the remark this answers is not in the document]
ezeike: If we remove the block request from the queue it will be sent again. We only want to remove it when there's an error or timeout. In those cases it is sent again.

It's also removed when the FSM processes it successfully. In that case the height is increased so the request won't be sent again.

controller/consensus.go
ezeike: We could do this and generally I prefer that approach but in this case there's very little to the queue.

One spot to add, two spots to remove, and it's not going to be used from anywhere else.

rem1niscence: @andrewnguyen22 thoughts on this? I honestly ultimately don't mind keeping it as-is but might be a little cleaner to have a custom type

Issue #154, opened 2025-04-21
[FEATURE] Convert `RootChainInfo` polling to Pub/Sub architecture


andrewnguyen22:
### Describe the feature you'd like
Convert `RootChainInfo` polling to a publish / subscribe architecture

### Motivation
RootChainInfo polling is unreliable and resource intensive
It should allow the `nestedChain` listeners to have the info quicker and more consistently

### Additional context
Have seen the following on Testnet and Alphanet:

- RootChain height can be off by 1 (N-1)
- RootChain updates can take > 1 second in worst case situations

### Possible implementation
Upgrade an http connection to websockets

#### Required Tags
- Priority: High
- Module: Multi-Module

PR #155, opened 2025-04-21
Issue #154

andrewnguyen22:
## Description
- RootChainInfo polling is unreliable and resource intensive
- It should allow the nestedChain listeners to have the info quicker and more consistently

Attempts to fix the following:

- RootChain height can be off by 1 (N-1)
- RootChain updates can take > 1 second in worst case situations

## Related Issues
Closes: #154 

## Changes Made
- Changed `RootChainInfo` retrieval from http polling to websockets pub/sub
- Converted `RootChainInfo` struct to protobuf for faster marshalling
- Still allow fallback http requests

## Checklist
- [x] I have run `npm run prettier` to format the web-wallet and/or block explorer (if applicable)

## Additional Notes
Still testing - will run in docker soon.


andrewnguyen22:
Currently broken - will address in the AM
image

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+	l             sync.Mutex                   // thread safety
+	afterRCUpdate func(info lib.RootChainInfo) // callback after the root chain info update
+	upgrader      websocket.Upgrader           // upgrade http connection to ws
+	log           lib.LoggerI                  // stout log
+}
+
+// NewRCManager() constructs a new instance of a RCManager
+func NewRCManager(controller *controller.Controller, c lib.Config, l lib.LoggerI) (man *RCManager) {
+	// create the manager
+	man = &RCManager{
```
rem1niscence, line 34: as we're aiming for understanding above all things I think that some of the naming, although it may seem obvious to most readers, is not clear at a glance ofwhat they do. (for example I was like "what is 'man'?" for a few seconds until I realized is short of manager). Expanding their name a bit could help with that 
```suggestion
func NewRCManager(controller *controller.Controller, config lib.Config, logger lib.LoggerI) (man *RCManager) {
	// create the manager
	manager = &RCManager{
```

cmd/rpc/server.go
```diff
@@ -120,8 +119,11 @@ func (s *Server) startRPC(router *httprouter.Router, port string) {
 	// Start RPC server
 	s.logger.Infof("Starting RPC server at 0.0.0.0:%s", port)
 	s.logger.Fatal((&http.Server{
-		Addr:    colon + port,
-		Handler: cor.Handler(http.TimeoutHandler(router, timeout, lib.ErrServerTimeout().Error())),
+		Addr:              colon + port,
+		ReadHeaderTimeout: timeout,
```
rem1niscence, line 123: having the same timeout for all operations may not be ideal if for example servers read heavier data than what they write, in practice it should be pretty similar but just as a possible suggestion the http timeouts in config could be separated as `httpReadTimeoutMS` and `httpWriteTimeoutMS` (the `ms` is to follow the same convention as the other timeouts in the config)

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+/* This file implements the client & server logic for the 'root-chain info' and corresponding 'on-demand' calls to the rpc */
+
+var _ lib.RCManagerI = new(RCManager)
+
+// RCManager handles a group of root-chain sock clients
+type RCManager struct {
+	c             lib.Config                   // the global node config
+	subscriptions map[uint64]*RCSubscription   // chainId -> subscription
+	subscribers   map[uint64]*RCSubscriber     // chainId -> subscriber
+	l             sync.Mutex                   // thread safety
```
rem1niscence, line 25: lock is being copied by value, that is not recommended and raise a flag in some linters (as it does on mine), please change this to a pointer reference 
```suggestion
	l             *sync.Mutex                   // thread safety
```

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+// NewRCManager() constructs a new instance of a RCManager
+func NewRCManager(controller *controller.Controller, c lib.Config, l lib.LoggerI) (man *RCManager) {
+	// create the manager
+	man = &RCManager{
+		c:             c,
+		subscriptions: make(map[uint64]*RCSubscription),
+		subscribers:   make(map[uint64]*RCSubscriber),
+		l:             controller.Mutex,
```
rem1niscence, line 38: why a mutex from the controller? from the commit I see it fixed the deadlock issue but is odd from a new user why would you use an external mutex, could you add a simple comment explaining why?

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+		afterRCUpdate: controller.UpdateRootChainInfo,
+		upgrader:      websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }},
+		log:           l,
+	}
+	// set the manager in the controller
+	controller.RCManager = man
```
rem1niscence, line 44: nitpick: This is more of a personal preference but I'm not a fan of the controller being modified here, seems like not something the `RCManager` should be in charge of as this is only concerned with performing the requests to get data from the root chain, no external modifications, this could be moved to `sever.go`'s `NewServer` instead imo. But feel free to disregard this if you do not agree

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+		// exit with 'not subscribed' error
+		return nil, lib.ErrNotSubscribed()
+	}
+	// get the info
+	info, err := sub.RootChainInfo(0, chainId)
+	if err != nil {
+		return nil, err
+	}
+	// update the info
+	sub.Info = *info
```
rem1niscence, line 119: generated rpc types contain internally a mutex too and this is also being copied by value which is not recommended `RCSubscription.Info` should be a pointer reference to avoid unexpected behavior with the lock

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+	books, err := sub.Orders(rootHeight, id)
+	// if an error occurred during the remote call
+	if err != nil {
+		// exit with error
+		return nil, err
+	}
+	// ensure the order book isn't empty
+	if books == nil || len(books.OrderBooks) == 0 {
+		// exit with error
+		return nil, lib.ErrEmptyOrderBook()
```
rem1niscence, line 173: should an empty order book really be an error? sounds like a normal situation for new or unused/abandoned chains

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+	// if the root chain id is the same as the info
+	sub, found := r.subscriptions[rootChainId]
+	if !found {
+		// exit with 'not subscribed' error
+		return nil, lib.ErrNotSubscribed()
+	}
+	// warn of the remote RPC call to the API of the 'root chain'
+	r.log.Warnf("Executing remote IsValidDoubleSigner call for address=%s at double sign height=%d and rootChainId=%d", address, height, rootChainId)
+	// exit with the results of the remote RPC call to the API of the 'root chain'
+	return sub.IsValidDoubleSigner(height, address)
```
rem1niscence, line 201: Here and in `GetCheckpoint` from the looks of the call there's no other way to retrieve the information besides doing a http request so a warning is given, for most of the other functions who do ultimately fall back to http when the data is not present it makes sense but this one seems like a false positive unless is expected for this to be converted to ws information too and if so it should be documented

rem1niscence (changes requested):
First set of review, most of them are minor clarifications and warnings, so far it looks pretty good! second one coming shortly

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+	s.rcManager.AddSubscriber(client)
+}
+
+// Add() adds the client to the manager
+func (r *RCManager) AddSubscriber(subscriber *RCSubscriber) {
+	// lock for thread safety
+	r.l.Lock()
+	defer r.l.Unlock()
+	// add to the map
+	r.subscribers[subscriber.chainId] = subscriber
```
rem1niscence, line 431: wouldn't subscribers mapped by chainID be always replaced if new ones come through? so only one subscriber per chainID could be kept, is that intentional?

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+	r.log.Errorf("WS Failed with err: %s", err.Error())
+	// close the connection
+	if err = r.conn.Close(); err != nil {
+		r.log.Error(err.Error())
+	}
+	// remove from the manager
+	r.manager.RemoveSubscriber(r.chainId)
+}
+
+const chainIdParamName = "chainId"
```
rem1niscence, line 455: is pretty odd to have a const at the bottom, please move it to the top of the file

controller/controller.go
```diff
@@ -29,10 +29,10 @@ type Controller struct {
 	Consensus *bft.BFT          // the async consensus process between the committee members for the chain
 	P2P       *p2p.P2P          // the P2P module the node uses to connect to the network
 
-	RootChainInfo lib.RootChainInfo // the latest information from the 'root chain'
-	isSyncing     *atomic.Bool      // is the chain currently being downloaded from peers
-	log           lib.LoggerI       // object for logging
-	sync.Mutex                      // mutex for thread safety
+	RCManager  lib.RCManagerI // the rpc client for the 'root chain'
```
rem1niscence, line 32: nitpick: I see the word `rpc` used multiple times to refer to the `RCManager` which under the hood uses a Websockets client which is based on `http` not `rpc`, is just a bit misleading unless you may refer to something else

cmd/rpc/sock.go
andrewnguyen22: It's related to https://github.com/canopy-network/canopy/issues/157

But ultimately, the pattern locking the `controller.Mutex` is the safest way to avoid multiple locks causing deadlocks.

cmd/rpc/sock.go
andrewnguyen22: I don't think it has a mutex

cmd/rpc/sock.go
andrewnguyen22: It doesn't have to be -- but that was previously the pattern

image

cmd/rpc/sock.go
andrewnguyen22: Hmmm this is a good point...

cmd/rpc/sock.go
```diff
@@ -0,0 +1,455 @@
...
+	r.l.Lock()
+	defer r.l.Unlock()
+	// add to the map
+	r.subscriptions[subscription.chainId] = subscription
+}
+
+// RemoveSubscription() gracefully deletes a RC subscription
+func (r *RCManager) RemoveSubscription(chainId uint64) {
+	// lock for thread safety
+	r.l.Lock()
```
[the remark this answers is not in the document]
andrewnguyen22: I think related to #157

cmd/rpc/sock.go
andrewnguyen22: Opened new issue #164

PR #193, opened 2025-05-22
fix: closed channel panic race conditon and update docker

pablocampogo:
- Fix panic on race condition of sends in closed channels.
- Update go version in docker

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.22.1-alpine AS builder
+FROM golang:1.23.0-alpine AS builder
 
 RUN apk update && apk add --no-cache make bash nodejs npm
 
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.23.0-alpine AS builder
+FROM golang:1.23.9-alpine AS builder
 
 RUN apk update && apk add --no-cache make bash nodejs npm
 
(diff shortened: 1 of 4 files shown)


p2p/p2p.go
```diff
@@ -219,25 +219,28 @@ func (p *P2P) Dial(address *lib.PeerAddress, disconnect, strictPublicKey bool) l
...
-	// always in case of error close connection
-	// first we need to check the error on creation to prevent panics
-	defer func() {
-		if err != nil {
-			connection.Stop()
-		}
-	}()
+	// in case of error close connection when flag is on
+	// not use closeOnFail if using retries
+	if closeOnFail {
```
rem1niscence, line 236: couldn't this be also done with sync.Once?

rem1niscence (commented):
just left a comment about a possible optimization

.docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.22.1-alpine AS builder
+FROM golang:1.23.0-alpine AS builder
```
rem1niscence, line 1: let's use the latest 1.23 image
```suggestion
FROM golang:1.23.9-alpine AS builder
```

p2p/p2p.go
pablocampogo: What we want is to not close it at all in some cases.

rem1niscence (approved):

PR #221, opened 2025-07-18
Fix for --data-dir and InitializeDataDirectory() call timing.

ezeike:
Cobra command line parameters are not yet parsed in the init() function, meaning the --data-dir parameter is not yet populated when InitializeDataDirectory is called.

This PR moves a few lines of code to the Start() function where the CLI parameters have been evaluated.

### DIFF
--- a/cmd/cli/cli.go
+++ b/cmd/cli/cli.go
@@ -29,6 +29,9 @@ import (
 var rootCmd = &cobra.Command{
 	Use:   "canopy",
 	Short: "the canopy blockchain software",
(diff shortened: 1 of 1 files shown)


pablocampogo (approved):

pablocampogo (changes requested):
I think we mandatorily need and env var

andrewnguyen22:
@ezeike ?

andrewnguyen22:
Closing for now. No response from author

PR #224, opened 2025-07-25
fix: useeffect -> queryAPI() function is fixed

coderprovider:
There are one issue in interval function of queryAPI
It reloads every 4 seconds, so when I move to other tab, I went back sometimes
I fixed useeffct and add the value to store the current tab and page


andrewnguyen22 (changes requested):
Did not fix the issue

https://github.com/user-attachments/assets/326d3cca-a151-428d-adcf-5ee6b5b9f9ec

andrewnguyen22:
Also this issue is for both the web wallet and block explorer... There was no code changes to the block explorer

PR #245, opened 2025-08-17
Issue #244

### LINKED ISSUE #244: [FEATURE] Add `RootBuildHeight` to the Quorum Certificate
### Describe the feature you'd like
Root build height should be included in the quorum certificate - not just be part of consensus BFT info

### Motivation
- Unlocks the ability for nested chains to utilize root chain information as part of their blocks as non-validators and historical syncers may use the RootBuildHeight

### Additional context
- Protocol (NestBFT) breaking: should have 3 phases 
  - 1) Backwards compatible feature add
  - 2) Upgrade height which switches to the new system once Validators are all upgraded
  - 3) Cleanup the deprecated code
- If this feature is used, checkpointing must be used in order to verify state after independence graduation

#### Required Tags
- Priority: High
- Module: BFT

andrewnguyen22:
closes #244

### DIFF
--- a/bft/README.md
+++ b/bft/README.md
@@ -136,6 +136,47 @@ Key aspects of Safety and Liveness Mechanisms:
 - Practical liveness guarantees under realistic network conditions
 - Resistance to various Byzantine attack vectors
 
+### Nested Problem A – Split at NEW_HEIGHT (Round 0)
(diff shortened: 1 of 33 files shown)


andrewnguyen22:
This is ready - but shouldn't be merged in until we're ready to execute the protocol breaking change (after my return October 9)

andrewnguyen22:
controller.CommitParallel doesn't play nice with current Eviction - should revert before merging

PR #250, opened 2025-08-29
Improve deletion and cleanup of DB data

rem1niscence:
## Description
Improve the deletion mechanism of keys to be deleted and adds a periodic "flusher" to force BadgerDB to delete stale/deleted data

## Related Issues

Closes: #<issue-number>

## Changes Made

- Add `Eviction` process to force deletion/flattening of stale data
- Fix all keys having a badger `noDiscardBit` preventing stale keys to be deleted
- Add periodic config for eviction process

## Checklist
- [ ] I have appropriately titled my branch `issue-#<issue-number>`.
- [x] I have run `npm run prettier` to format the web-wallet and/or block explorer (if applicable)

## Additional Notes

This only deletes keys using this version onwards, keys set beforehand wont be deleted, a full chain resync or snapshot from syncing this branch is recommended


store/store.go
```diff
@@ -194,6 +196,15 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
 	if e := s.writer.Flush(); e != nil {
 		return nil, ErrCommitDB(e)
 	}
+	// Trigger eviction of LSS deleted keys
+	// check if the current version is a multiple of the cleanup block interval
+	if s.Version()%s.config.StoreConfig.CleanupBlockInterval == 0 {
```
andrewnguyen22, line 201: Should we have a default set if this config is 0?

store/store.go
```diff
@@ -364,6 +375,112 @@ func (s *Store) setCommitID(version uint64, root []byte) lib.ErrorI {
...
+	triggerEvict := fmt.Appendf(nil, "zzz/trigger_key_%s", randValue())
+	writer := s.db.NewWriteBatchAt(lssVersion)
+	if err := writer.Set(triggerEvict, nil); err != nil {
+		return ErrStoreSet(err)
+	}
+	if err := writer.Flush(); err != nil {
+		return ErrCommitDB(err)
+	}
+	// this is basically a hack to force deletion of entries marked for eviction
+	// mark evict with the correct meta bits
```
andrewnguyen22, line 399: Comment cleanup? Let's not use the word 'hack' let's say 'work around'

andrewnguyen22 (changes requested):

store/store.go
```diff
@@ -194,6 +196,15 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
 	if e := s.writer.Flush(); e != nil {
 		return nil, ErrCommitDB(e)
 	}
+	// Trigger eviction of LSS deleted keys
+	// check if the current version is a multiple of the cleanup block interval
+	if s.Version()%s.config.StoreConfig.CleanupBlockInterval == 0 {
+		go func() {
+			if err := s.Evict(); err != nil {
```
andrewnguyen22, line 203: Is this function triggered anywhere else besides startup?

andrewnguyen22:
@rem1niscence reviewed your PR - looks good but it's unclear to me at this time if you are triggering Evict regularly or not... Also, should we regularly run 'GarbageCollection' to prevent these node kills we've seen?

store/store.go
rem1niscence: no, is checked after every commit and runs every X configured blocks (default 200)

store/store.go
rem1niscence: you're right but that should be addressed on the config.go where is being set, will perfom the change there

store/store.go
```diff
@@ -364,6 +375,112 @@ func (s *Store) setCommitID(version uint64, root []byte) lib.ErrorI {
...
+		return ErrCommitDB(err)
+	}
+	// reset discard timestamp after eviction
+	defer s.db.SetDiscardTs(0)
+	// flatten the DB to optimize the storage layout
+	if err := s.db.Flatten(runtime.NumCPU() / 2); err != nil {
+		return ErrCommitDB(err)
+	}
+	// log the results
+	total, deleted := s.keyCount(lssVersion, []byte(latestStatePrefix), true)
```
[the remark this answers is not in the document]
rem1niscence: I changed it to DEBUG, I think is valuable to know the amount of data set there and right now, the numbers never go past 2k tops, we may remove it if the number is bigger but is a good way to know how big our lss is getting

rem1niscence:
Addresses comments, also added the GC with the default discard size as I had it previously but I haven't completely benchmarked how much it affects DB operations while it runs. That number may change in the future

andrewnguyen22 (approved):

Issue #241, opened 2025-08-10
[FEATURE] Implement AMM Spec


andrewnguyen22:
## dex.go - implements AMM style atomic swaps between root & nested chains
--
## Overview

This file is designed to facilitate a Uniswap V2 style AMM token swap
between a root and a nested chain.

The AMM executes **batched limit orders** and supports **one-sided liquidity deposits and withdrawals**.

Batches are processed **atomically between chains** via receipts and a strict locking protocol, ensuring deterministic outcomes and correct LP point allocation.

## 1. Core Idea: Uniswap AMM formula

At its heart, the Uniswap invariant is:

```
x * y = k
```

- `x` = reserve of token A (add-side)
- `y` = reserve of token B (distribute-side)
- `k` = constant product

In `dex.go`:

```go
dY = (amountInWithFee * y) / (x*1000 + amountInWithFee)
```

This is literally the **Uniswap V2 formula** with fee `0.3%`. Every swap still satisfies `x * y` invariant (ignoring rounding errors).

---

## 2. Virtual Pool Mirrors Counterparty State

- Each chain maintains a `virtualPool` based on the **other chain’s last `PoolSize`**.
- When processing orders, it treats this as the effective reserve (`x`) for the “add” token.
- Trades are applied **against local liquidity (`y`)**, but the invariant is computed against the mirrored counterparty pool.

---

## 3. LP Points Are Standard Pro-Rata

Liquidity addition/withdrawal follows the classic Uniswap V2 integer math:

```go
ΔL = L * (√((x+d)*y) - √(x*y)) / √(x*y)
```

- LP points proportional to **square root of the pool product** (`√k`)
- One-sided deposits work because counterparty’s pool size is included (`x` = remote PoolSize)
- Withdrawals are proportional to **LP share**

---

## 4. Batches Don’t Break Uniswap Logic

- Batches aggregate multiple orders → atomic execution, prevent reentrancy, ensure deterministic receipts.
- Within a batch:
    - Orders processed sequentially (pseudorandomly to prevent front-running)
    - Each trade updates `x` and `y` (virtual and local reserves)
    - `dY` computation is identical to Uniswap

[x] Batching = different implementation, same economic outcome.

---

## 5. Cross-Chain Receipts Ensure Determinism

- Nested reads Root’s locked batch and vice versa.
- Trades only finalize after receipts → both sides see the same LP state.
- Failed orders refunded → invariant preserved.

---

## 6. Fees Are Applied Identically

```go
[...]

PR #254, opened 2025-09-10
Add events indexer

pablocampogo:


andrewnguyen22:
Not implemented correctly. Merging into #241 to fix

PR #275, opened 2025-11-14
fix: add reflect check to prevent panic

pablocampogo:
Add reflect field check as bandaid to prevent nil pointer deference in pebble

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -520,18 +520,15 @@ func (s *Store) Compact(version uint64, compactHSS bool) lib.ErrorI {
 	// check if batch's db closed field is not nil using reflection
 	batchValue := reflect.ValueOf(batch).Elem()
 	batchDBField := batchValue.FieldByName("db")
-	if batchDBField.IsValid() && !batchDBField.IsNil() {
-		dbValue := batchDBField.Elem()
-		closedField := dbValue.FieldByName("closed")
-		if closedField.IsNil() {
-			s.mu.Unlock() // unlock commit op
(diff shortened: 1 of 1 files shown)


store/store.go
```diff
@@ -516,6 +517,22 @@ func (s *Store) Compact(version uint64, compactHSS bool) lib.ErrorI {
...
+		if closedField.IsNil() {
+			s.mu.Unlock() // unlock commit op
+			s.log.Debugf("key compaction skipped [%d]: closed field is nil", version)
+			return nil
+		}
+	} else {
+		s.mu.Unlock() // unlock commit op
+		s.log.Debugf("key compaction skipped [%d]: db field is nil", version)
+		return nil
+	}
```
rem1niscence:
```suggestion
	batchValue := reflect.ValueOf(batch).Elem()
	batchDBField := batchValue.FieldByName("db")
	if !batchDBField.IsValid() || batchDBField.IsNil() {
		s.mu.Unlock() // unlock commit op
		return ErrCommitDB(fmt.Errorf("db field is nil"))
	}
	dbValue := batchDBField.Elem()
	if closedField := dbValue.FieldByName("closed"); closedField.IsNil() {
		s.mu.Unlock() // unlock commit op
		return ErrCommitDB(fmt.Errorf("closed field is nil"))
	}
```
This way it uses early returns and avoids nested segments, while also leaving the caller to do the error log

rem1niscence (approved):
just a minor refactor suggestion

PR #279, opened 2025-11-21
chore: change bft logs to debug and allow non start calls

pablocampogo:
- Change majority votes logs to debug
- Unbreak config (when the canopy code is called without a start) so the setup.sh of the deployments repo works
- Delete dockerfile from root since it is broken


cmd/auto-update/main.go
```diff
@@ -39,12 +38,13 @@ var (
...
 func main() {
-	// check if start was called
-	if len(os.Args) < 2 || os.Args[1] != "start" {
-		log.Fatalf("invalid input, only `start` command is allowed")
-	}
 	// get configs and logger
 	configs, logger := getConfigs()
+	// check if no start was called, this means it was just called as config
+	if len(os.Args) < 2 || os.Args[1] != "start" {
+		logger.Info("config completed")
```
rem1niscence: this is misleading, custom setups (like mine and anybody else not using the deployments repo now or in the future) will see this and be confused about why is doing nothing when the command used is not `start`, the log should provide more information about what happened and let the users know what to do to run other commands (like using the other binary for canopy commands)

PR #292, opened 2026-01-28
Add plugins support in auto update

pablocampogo:
Add plugins support in auto update


rem1niscence (changes requested):
A refactor of the autoupdater code is required in order for this to be scalable and merged for production usage, in the comments I left multiple suggestions on how to do it but basically it resumes into 
1. not creating specialized structs `PluginRelease` but augment the ones that already exists `Release`
2. Do not create special logic for plugin lifetime on the coordinator but reuse the `Supervisor` to manage the plugins through it, and the coordinator will simply need to loop through them and update as needed
3. This is not on the comments but also required: Right now the plugin update and canopy release update are done together, that is dangerous and could leave the system on a broken state if any of the plugin fails for whatever reasone, the canopy binary must always run so the plugin lifetime/update must be done separate from the canopy binary

cmd/auto-update/updater.go
```diff
@@ -26,163 +26,246 @@ type GithubRelease struct {
...
+		arch := runtime.GOARCH
+		if arch == "amd64" {
+			arch = "x64"
+		}
+		return fmt.Sprintf("csharp-plugin-%s-%s.tar.gz", runtime.GOOS, arch)
+	case "python":
+		return "python-plugin.tar.gz"
+	default:
+		return fmt.Sprintf("%s-plugin.tar.gz", rm.config.PluginType)
+	}
```
rem1niscence: let's remove the hardcoded values from this struct, make them a map in `ReleaseManager`, type could be `map[string]PluginConfig` with fields like `Name` `FileName`, and whatever is needed, then, write that map on `cmd/auto-update/main.go` along with the other config of the `ReleaseManager`. That gives the advantage of

1) Removing hardcoded value from here
2) Making the `ReleaseManager` agnostic, it does not care about specific plugins
3) If a new plugin is needed, simply add them to the existing map on `main.go` without needing to come here.

after that, `getAssetName` would only iterate that map looking for the `FileName` on this case and a `isX64` flag for example that will be applied only to C# (or any other that may also need it). Making this way simpler, flexible and agnostic

cmd/auto-update/coordinator.go
```diff
@@ -278,30 +343,25 @@ func (c *Coordinator) ApplyUpdate(ctx context.Context, release *Release) error {
...
-		timer := time.NewTimer(delay)
-		// allow cancellation of timer if context is done
-		defer timer.Stop()
-		select {
-		case <-ctx.Done():
-			return ctx.Err()
-		case <-timer.C:
-		}
-	}
+
```
rem1niscence: why the random delay for staggered updates was removed? it is needed

rem1niscence (commented):
The main takeaway from this review is to not have any hardcoded values in both `coordinator.go` and `updater.go`. Add those values to the `main.go` and pass them as configs

rem1niscence (changes requested):
code is getting better, keep up the good work

rem1niscence (changes requested):

rem1niscence (approved):

wanghaitao666777888 (commented):
dfhgd

daino0:
> Add plugins support in auto update

The official go implementation of the Canopy Network protocol txn and explain boss how to creat viedo tutorial

PR #315, opened 2026-02-19
feat: new buyers filter index for orders

pablocampogo:
Add buyerSendAddress index and handle when buyer changes

### DIFF
--- a/fsm/swap.go
+++ b/fsm/swap.go
@@ -263,20 +263,9 @@ func (s *StateMachine) CloseOrder(orderId []byte, chainId uint64) (err lib.Error
 
 // SetOrder() sets the sell order in state
 func (s *StateMachine) SetOrder(order *lib.SellOrder, chainId uint64) (err lib.ErrorI) {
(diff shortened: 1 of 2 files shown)


fsm/swap.go
```diff
@@ -262,6 +263,21 @@ func (s *StateMachine) CloseOrder(orderId []byte, chainId uint64) (err lib.Error
...
+	}
+	// if existing order has a buyer and it differs from the new order's buyer, clean up the old index
+	if existingOrder != nil && len(existingOrder.BuyerSendAddress) > 0 {
+		// remove old buyer index if buyer changed or was removed
+		if !bytes.Equal(existingOrder.BuyerSendAddress, order.BuyerSendAddress) {
+			if err = s.Delete(KeyForOrderByBuyer(existingOrder.BuyerSendAddress, chainId, order.Id)); err != nil {
+				return
+			}
+		}
+	}
```
rem1niscence: This cleanup logic is not completely related to set order and is slightly complex (3 nested ifs) removed to an external function and call it as needed

PR #316, opened 2026-02-20
feat: make addresses indexes for orders optional

pablocampogo:
## Summary

- Added `IndexOrdersByAddresses` configuration option to `StateMachineConfig` that controls whether secondary indexes for order lookups by seller/buyer address are created
- When disabled, the `/v1/query/orders` endpoint returns an error if `sellersSendAddress` or `buyerSendAddress` filters are used
- Default value is `true` (indexes enabled)

### Configuration (`lib/config.go`)
- Added `DefaultIndexOrdersByAddresses` constant
- Added `IndexOrdersByAddresses` field to `StateMachineConfig`

### State Machine (`fsm/swap.go`)
- `SetOrder`: Conditionally creates seller/buyer secondary indexes based on config
- `cleanupStaleBuyerIndex`: Skips cleanup when indexing is disabled
- `DeleteOrder`: Conditionally deletes seller/buyer secondary indexes based on config

### RPC Endpoint (`cmd/rpc/query.go`)
- `Orders`: Returns descriptive error when address-based filtering is requested but indexing is disabled

### DIFF
--- a/cmd/rpc/query.go
+++ b/cmd/rpc/query.go
@@ -284,10 +284,19 @@ func (s *Server) Order(w http.ResponseWriter, r *http.Request, _ httprouter.Para
 // Orders retrieves the order book for a committee with optional filters and pagination
 func (s *Server) Orders(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	// Invoke helper with the HTTP request, response writer and an inline callback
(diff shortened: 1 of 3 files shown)


pablocampogo:
These indexes will be deleted over all

PR #394, opened 2026-05-12
feat(cli): add --rpc-url/--admin-url flags and fix --data-dir targeting

ezeike:
## Summary

- Fix `--data-dir` flag being ignored: moved initialization from `init()` to `PersistentPreRunE` so flags are parsed before config loads
- Add `--rpc-url` and `--admin-url` flags to override node URLs from config at runtime

## Test plan

- [ ] `canopy --data-dir /custom/path <cmd>` uses the specified directory
- [ ] `canopy --rpc-url http://localhost:9000 <cmd>` overrides config RPC URL
- [ ] `canopy --admin-url http://localhost:9001 <cmd>` overrides config admin URL
- [ ] Default behavior unchanged when flags are omitted

### DIFF
--- a/cmd/cli/cli.go
+++ b/cmd/cli/cli.go
@@ -29,6 +29,16 @@ import (
 var rootCmd = &cobra.Command{
 	Use:   "canopy",
 	Short: "the canopy blockchain software",
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


pablocampogo:
PR will be closed, fixes and docker and auto update usage addressed here https://github.com/canopy-network/canopy/pull/512

PR #402, opened 2026-05-15
Added Low Spec Hardware Optimization Guide

sniperwolf260879:
I've been stress-testing a validator node on legacy hardware (AMD FX-6300). This PR adds a "Low Spec Optimization Guide" to help users maintain uptime on older builds, covering BIOS, SSD, and OS stability.


rem1niscence:
thank you for this but this kind of information is better suited to be on our validator docs rather than in the repository as node runners are more likely to check there, I will close this but pass this to our team in order to add this information to the validator docs in the near future

PR #403, opened 2026-05-18
feat: implement block subscription and enhance polling mechanisms for improved dashboard updates

XJuanCarlosXD:
The explorer dashboard now auto-refreshes whenever a new block arrives, and the console is no longer flooded with errors when the RPC node is degraded. A new global block subscription detects when the chain height advances and refreshes the entire dashboard at once, so blocks, transactions, validators, and stats stay in sync without needing a manual page reload. 

All RPC traffic is routed through a same-origin proxy in both production and local development, which fixes the CORS errors that appeared on deploys other than the main domain and on localhost. The blocks query was simplified to a single request per cycle instead of ten parallel ones, and all polling queries now back off automatically when the upstream node fails so the dashboard stops hammering a node that is returning 502s. The check interval was lowered to 1 second, and all internal console.log noise was removed.

@rem1niscence

### DIFF
--- a/cmd/rpc/web/explorer/env.example
+++ b/cmd/rpc/web/explorer/env.example
@@ -18,3 +18,13 @@ VITE_PUBLIC_ADMIN_RPC_URL=https://node1.canopy.us.nodefleet.net/admin/
 
 # Environment mode (development, production)
 VITE_NODE_ENV=development
+
+# Block polling interval in ms (default 2000, minimum 500).
+# A simple explorer/wallet or private deployment can keep the 2s default to
+# reduce node load; lower it (e.g. 1000) for snappier dashboard updates.
+# VITE_BLOCKS_POLL_MS=2000
+
+# Dev-only: upstream targets for the Vite proxy (/rpc-node1, /admin-node1).
+# Defaults to VITE_PUBLIC_RPC_URL / VITE_PUBLIC_ADMIN_RPC_URL when unset.
+# VITE_DEV_RPC_PROXY=http://localhost:50002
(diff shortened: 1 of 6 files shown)


rem1niscence:
not sure whether a default of 1 second is a bit too much as if multiple clients are connected to a validator and running at the same time they could affect consensus as some consensus calls are blocked by rpc requests (is something that is pending to be fixed), can it be lowered by at least 2 seconds?

XJuanCarlosXD:
for now. I'm always looking for a change in getblock every 1 sec by using getblock and if there's a change I extract all blocks for the explorer. in our case there's no issue with our infra. but for other users that also won't affect them since it's only 1 getblock. but in any case I can create a custom variable named "EXPLORER_SYNC_TIME=2" which by default could be 2 for other use cases AND valitators but we need ours faster since it's used by more people. @rem1niscence

rem1niscence:
then on that case do that, default it to two and in your specific config change it to 1 as it should be configurable as you mentioned. That way a default explorer/wallet does not need to poll that often, specially helpful for simple private deployments

XJuanCarlosXD:
Made the block polling interval configurable via the VITE_BLOCKS_POLL_MS env var. The default is now 2s so a plain explorer/wallet — especially a simple private deployment — doesn't poll the node aggressively; it can be overridden per environment for snappier updates (values below the 500ms floor or invalid ones fall back to the default). BLOCKS_POLL_MS drives useLatestBlock and useCardData, which is what triggers new-block detection in useBlockSubscription.  @rem1niscence

rem1niscence (approved):

PR #404, opened 2026-05-21
vesting dao mint

andrewnguyen22:
## Summary

  This PR adds vesting-aware account support and a mint option for DAO transfer proposals.

  ## What Changed

  ### Vesting accounts

  - Added vesting metadata to accounts:
      - vestingAmount
      - vestingStartHeight
      - vestingCliffHeight
      - vestingEndHeight
  - Added vesting-aware send support so a transfer can create or extend a recipient vesting schedule.
  - Enforced spendable-balance checks in the FSM so locked vesting funds cannot be spent.
  - Updated RPC/account JSON handling so vesting fields are preserved in responses and serialization.
  - Fixed faucet top-up behavior to use spendable balance rather than raw total balance.

  ### DAO transfer minting

  - Added mint: bool to MessageDAOTransfer.
  - When mint=true, the FSM mints amount into the DAO pool immediately before executing the DAO transfer.
  - This allows validator-approved one-off treasury mint events for grants.

  ## Why

  ### Vesting

  The chain needs to distinguish between:

  - total account balance
  - vesting-restricted balance
  - currently spendable balance

  This PR keeps total ownership at the storage layer while enforcing spendability dynamically based on vesting schedule and height.

  ### DAO minting

  This gives validators a controlled way to approve exceptional treasury grant events that require new issuance, without creating a separate
  governance message type.

  ## Testing

  Covered with focused tests for:

  - vesting spendable accounting
  - vesting JSON round-trip preservation
  - RPC account responses with spendable amount and vesting metadata
  - DAO transfer behavior
  - DAO mint-then-transfer behavior

### DIFF
--- a/cmd/cli/cli.go
+++ b/cmd/cli/cli.go
@@ -3,7 +3,6 @@ package cli
 import (
 	"encoding/json"
 	"errors"
-	"flag"
 	"fmt"
 	"log"
 	"os"
@@ -45,7 +44,6 @@ var (
 )
 
 func init() {
-	flag.Parse()
 	rootCmd.AddCommand(startCmd)
 	rootCmd.AddCommand(versionCmd)
 	rootCmd.AddCommand(queryCmd)
--- a/cmd/rpc/query.go
+++ b/cmd/rpc/query.go
@@ -739,13 +739,26 @@ func (s *Server) heightAndAddressParams(w http.ResponseWriter, r *http.Request, 
 	})
 }
 
-func spendableAccountView(sm *fsm.StateMachine, account *fsm.Account) *fsm.Account {
+func spendableAccountView(sm *fsm.StateMachine, account *fsm.Account) *AccountView {
 	if account == nil {
 		return nil
 	}
-	view := *account
-	view.Amount = sm.AccountSpendableAmount(account)
-	return &view
+	total := account.Amount
+	spendable := sm.AccountSpendableAmount(account)
+	vested := sm.AccountVestedAmount(account)
+	locked := sm.AccountLockedAmount(account)
+	return &AccountView{
+		Address:            account.Address,
+		Amount:             spendable,
+		TotalAmount:        total,
+		SpendableAmount:    spendable,
+		VestedAmount:       vested,
+		LockedAmount:       locked,
+		VestingAmount:      account.VestingAmount,
+		VestingStartHeight: account.VestingStartHeight,
+		VestingCliffHeight: account.VestingCliffHeight,
+		VestingEndHeight:   account.VestingEndHeight,
+	}
 }
 
 func spendableAccountPageView(sm *fsm.StateMachine, page *lib.Page) *lib.Page {
@@ -757,7 +770,7 @@ func spendableAccountPageView(sm *fsm.StateMachine, page *lib.Page) *lib.Page {
 	if !ok || accounts == nil {
 		return &view
 	}
-	result := make(fsm.AccountPage, 0, len(*accounts))
+	result := make(AccountViewPage, 0, len(*accounts))
 	for _, account := range *accounts {
 		result = append(result, spendableAccountView(sm, account))
 	}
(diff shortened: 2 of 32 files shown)


rem1niscence:
is it ready? @andrewnguyen22 as some tests are failing on the CI

fsm/message.go
```diff
@@ -54,10 +54,20 @@ func (s *StateMachine) HandleMessage(msg lib.MessageI) lib.ErrorI {
 
 // HandleMessageSend() is the proper handler for a `Send` message
 func (s *StateMachine) HandleMessageSend(msg *MessageSend) lib.ErrorI {
+	// vesting sends must preflight recipient compatibility before mutating sender state
+	if msg.VestingStartHeight != 0 || msg.VestingEndHeight != 0 {
```
pablocampogo, line 58: There should be a validation that x.VestingStartHeight is not lower than currentHeight

fsm/account.go
```diff
@@ -152,6 +211,52 @@ func (s *StateMachine) AccountAdd(address crypto.AddressI, amountToAdd uint64) l
...
+	acc, err := s.GetAccount(crypto.NewAddress(msg.ToAddress))
+	if err != nil {
+		return err
+	}
+	// overflow protection
+	if acc.Amount > math.MaxUint64-msg.Amount || acc.VestingAmount > math.MaxUint64-msg.Amount {
+		return ErrInvalidAmount()
+	}
+	// update amount
+	acc.Amount += msg.Amount
```
pablocampogo: Since s.GetAccount returns a pointer that changes cache values this mutation should happen after validation

pablocampogo (changes requested):

fsm/message.go
andrewnguyen22: This is actually intentional - to allow us to retroactively set vesting terms.

andrewnguyen22:
Review re-requested @rem1niscence @pablocampogo @ezeike

cmd/cli/cli.go
```diff
@@ -45,7 +44,6 @@ var (
 )
 
 func init() {
-	flag.Parse()
```
rem1niscence, line 48: just to make sure, did you remove this because cobra is being used here? this won't any any other command right?

fsm/account.go
```diff
@@ -81,8 +81,67 @@ func (s *StateMachine) GetAccountBalance(address crypto.AddressI) (uint64, lib.E
...
+		return 0
+	}
+	// get the amount that is currently ineligible for spend
+	locked := s.AccountLockedAmount(account)
+	// check for invariant
+	if locked >= account.Amount {
+		return 0
+	}
+	// return spendable amount
+	return account.Amount - locked
```
rem1niscence, line 138:
```suggestion
	// return spendable amount
	return max(0, account.Amount - s.AccountLockedAmount(account))
```
minor refactor to simplify the check, other functions like `AccountLockedAmount` could have something similar but this could be objected in favour of being explicit on what's happening

cmd/rpc/query.go
```diff
@@ -731,6 +739,32 @@ func (s *Server) heightAndAddressParams(w http.ResponseWriter, r *http.Request,
 	})
 }
 
+func spendableAccountView(sm *fsm.StateMachine, account *fsm.Account) *fsm.Account {
+	if account == nil {
+		return nil
+	}
+	view := *account
```
rem1niscence: warning on my linter here: 

```
assignment copies lock value to view: github.com/canopy-network/canopy/fsm.Account contains google.golang.org/protobuf/runtime/protoimpl.MessageState contains sync.Mutex (copylocks default)
```

rem1niscence (changes requested):
everything looks all right in general, just set it as `request changes` to enforce the response of the comments. Also in general, I'm pretty sure you did but just to double confirm, as there's a possibly high amount of places where account balances are manipulated in general, did you make sure that all these places have the checks for vesting in place? wether to view those accounts or manipulate them in any way

fsm/account.go
andrewnguyen22: I think explicit is better here - cause the 0 can be confusing, also you might be subject to overflow on that implementation

cmd/rpc/query.go
andrewnguyen22: On it

cmd/cli/cli.go
andrewnguyen22: Yes - flag.Parse() was removed because this CLI is using Cobra flags, not the standard library flag package, and parsing stdlib flags in init() was both unnecessary and potentially wrong.

rem1niscence (approved):

PR #408, opened 2026-06-03
Fix smt pebble

pablocampogo:


cmd/rpc/sock.go
```diff
@@ -442,8 +442,12 @@ func (r *RCSubscription) Listen() {
 		r.manager.l.Lock()
 		// update the root chain info
 		r.Info = newInfo
-		// execute the callback
-		r.manager.afterRCUpdate(newInfo)
+		// skip the callback during syncing: the BFT is paused so ResetBFT messages are
+		// unnecessary, and UpdateP2PMustConnect sends to a channel with buffer=1 that can
+		// block while holding the controller lock, starving the sync loop
```
rem1niscence:
```suggestion
		// skip the callback during syncing: the BFT is paused so ResetBFT messages are
		// unnecessary
```
over explaining the other part of a module on the controller, could be more confusing than helpful, specially in the future when the context of it is gone, better to keep the comments helpful but concrete

store/store.go
```diff
@@ -472,6 +473,39 @@ func (s *Store) IncreaseVersion() { func() { s.version++; s.sc = nil }() }
...
 func (s *Store) Version() uint64 { return s.version }
 
+// SetSyncing tells the store whether the node is currently syncing, allowing it to
+// defer expensive maintenance operations (compaction) that cause write stalls at scale.
+func (s *Store) SetSyncing(v bool) { s.syncing.Store(v) }
+
+// CompactAll runs compaction across all store prefixes including SMT commitment nodes
+// and indexer entries that are not covered by the regular MaybeCompact cycle.
+// Should be called after sync completes to consolidate the accumulated versions.
+func (s *Store) CompactAll() {
```
rem1niscence: This function duplicates the work of the `Compact` function in this same file, this add 2 different ways of performing the same compaction even with intertwining prefixes (both compact lss and hss), the `Compact` function already has the capabilities of this, but it could be refactor to a helper `compactPrefixRange` where a prefix is passed and it compacts that, if you want to perform this (which is a good idea), please remove `CompactAll` completely, and refactor the compact function as mentioned, that way you could pass an arg `synced` or something like that which will trigger the additional compactions of the `stateCommitmentPrefix`, `stateCommitIDPrefix` and `indexerPrefix`.

store/store.go
```diff
@@ -474,6 +473,39 @@ func (s *Store) IncreaseVersion() { func() { s.version++; s.sc = nil }() }
 // number of the state. This is used to track the versioning of the state data.
 func (s *Store) Version() uint64 { return s.version }
 
+// SetSyncing tells the store whether the node is currently syncing, allowing it to
+// defer expensive maintenance operations (compaction) that cause write stalls at scale.
```
rem1niscence:
```suggestion
// SetSyncing tells the store whether the node is currently syncing
```
Extraneous comment portion, at the method declaration a comment should explain **what** it does, leave the **why** to the caller as otherwise you're implying this can only be used for that which should not be the case

store/versioned_store.go
```diff
@@ -77,6 +77,18 @@ func NewVersionedStore(db pebble.Reader, batch *pebble.Batch, version uint64) *V
 	}
 }
 
+// NewParallelReader creates a read-only VersionedStore sharing the same
+// underlying database (snapshot) but with its own buffers, safe for concurrent
+// use from a separate goroutine. The returned store must NOT be closed, as it
+// does not own the underlying reader.
+func (vs *VersionedStore) NewParallelReader() *VersionedStore {
```
rem1niscence, line 85: This is incomplete, do not leave the task of not closing the reader to the user, this could lead to production bugs, add a field on the `VersionedStore` to signify whether is parallel or not and if `Closed` is called with that set to true, either produce a no-op or panic. I'd rather panic to ensure the user that must not be done

store/store.go
```diff
@@ -792,6 +767,24 @@ func (s *Store) Compact(version uint64, compactHSS bool) lib.ErrorI {
 		// update HSS compaction metrics
 		s.metrics.UpdateStoreJobMetrics(0, hssDuration, 0)
 	}
+	// post-sync compaction: consolidate the prefixes excluded from the regular compaction cycle
+	// (SMT commitment nodes, commit ids, and indexer entries)
+	if synced {
```
rem1niscence: compaction should not be aware of syncing, if you see on the arguments, one of them is `compactHSS` which signals whether the HSS should be compacted, do the same for the other IDs, that way is generic for whether the called wants to compact a given prefix or not

store/store.go
```diff
@@ -759,31 +723,42 @@ func (s *Store) MaybeBackup() {
...
 }
 
-// Compact runs Pebble range compaction over the latest and optional historic state prefixes.
-func (s *Store) Compact(version uint64, compactHSS bool) lib.ErrorI {
-	// first compaction: latest state  keys
-	startPrefix, endPrefix := latestStatePrefix, prefixEnd(latestStatePrefix)
+// Compact runs Pebble range compaction over the store prefixes. It always compacts the latest
+// state prefix, optionally the historic state prefix when 'compactHSS' is set, and—when 'synced'
+// is set (post-sync consolidation)—the state commitment, commit id, and indexer prefixes which
+// are not touched by the regular compaction cycle.
```
rem1niscence: LLM excessive comment, please reduce it to what's important

store/store.go
```diff
@@ -759,31 +723,42 @@ func (s *Store) MaybeBackup() {
...
+		return nil
+	}
+	defer s.compaction.Store(false)
 	// track current time and version
 	now := time.Now()
 	s.log.Debugf("key compaction started at height %d", version)
-	// create a timeout to limit the duration of the compaction process
-	// TODO: timeout was chosen arbitrarily, should update the number once multiple tests are run
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	// create a timeout to limit the duration of the compaction process; a full post-sync
```
rem1niscence: remove information about sync, this method should not care about syncing at all

store/versioned_store.go
```diff
@@ -183,6 +185,12 @@ func (vs *VersionedStore) Commit() (e lib.ErrorI) {
 
 // Close closes the store and releases resources
 func (vs *VersionedStore) Close() lib.ErrorI {
+	// a parallel reader shares (does not own) the underlying reader; closing it would
+	// release a snapshot still in use by the owner. This is always a programming error
+	// so panic to surface it rather than silently corrupting concurrent readers.
+	if vs.parallel {
+		panic("Close() called on a parallel VersionedStore reader: parallel readers do not own the underlying reader and must never be closed")
```
rem1niscence: simplify panic message

rem1niscence (changes requested):

store/smt.go
```diff
@@ -189,49 +188,102 @@ func (s *SMT) Commit(unsortedOps map[uint64]valueOp) (err lib.ErrorI) {
...
-	// add 16 synthetic borders to the tree
-	cleanup, err := s.addSyntheticBorders()
+	// fall back to sequential processing when operations are fewer than subtrees;
+	// addSyntheticBorders + cleanup overhead dominates for small batches
+	if len(unsortedOps) < NumSubtrees*2 {
+		return s.Commit(unsortedOps)
+	}
+
+	parentTxn, ok := s.store.(*Txn)
+	if !ok {
```
rem1niscence, line 206: add a comment for this case

store/smt.go
```diff
@@ -189,49 +188,102 @@ func (s *SMT) Commit(unsortedOps map[uint64]valueOp) (err lib.ErrorI) {
...
+		return s.Commit(unsortedOps)
+	}
+
+	groups, err := s.sortOperationsByPrefix(unsortedOps)
```
rem1niscence, line 211: add a comment of what's going on here

rem1niscence (approved):

PR #411, opened 2026-06-13
Eth rpc indexer staging pr

andrewnguyen22:
## Summary

  This PR tightens the ETH RPC translation layer so Canopy behaves more like an EVM-compatible chain on the transaction lifecycle and nonce paths that CEXs and Ethereum wallets depend on.

  The focus is on:
  - canonical Ethereum tx hash behavior
  - canonical tx/receipt/log response shapes
  - safer and more predictable `eth_getTransactionCount`
  - better pending/mined lookup continuity
  - smaller indexer support for mined Ethereum nonce tracking

  ## What changed

  ### ETH transaction / receipt compatibility
  - Return the canonical Ethereum tx hash from `eth_sendRawTransaction`
  - Use that same hash consistently across:
    - `eth_getTransactionByHash`
    - `eth_getTransactionReceipt`
    - block-embedded full transaction responses
    - supported log responses
  - Separate transaction and receipt response types
  - Return canonical receipt log objects instead of string arrays
  - Return `null` instead of RPC errors for unknown tx lookups where Ethereum clients expect `null`

  ### Nonce handling
  - Rework `eth_getTransactionCount` for Ethereum-derived accounts to use:
    - latest mined Ethereum nonce from the indexer
    - plus local pending state for `"pending"`
  - Keep the call side-effect free
  - Align nonce/read behavior with Canopy replay-protection bounds
  - Reject saturated `"pending"` windows instead of returning an already-occupied nonce
  - Restore immediate same-sender same-nonce replacement cleanup in the local pending cache

  ### Indexer support
  - Persist the latest mined Ethereum nonce per sender for RLP-backed transactions
  - Recompute or delete that nonce anchor correctly on rollback/delete paths
  - Index RLP-backed transactions by canonical Ethereum tx hash as an alias of the stored Canopy tx result
  - Clean up sender/recipient and Ethereum-hash alias index rows on delete paths

  ### Ethereum RPC behavior improvements
  - Improve block-tag handling on `...ByNumber` methods
  - Tighten `eth_getLogs` filtering behavior for the supported pseudo-token surface
  - Add `eth_protocolVersion`
  - Improve some wallet-facing tx/block response fields
  - Add JSON-RPC batch request handling

  ### Docs
  - Update ETH RPC docs to clarify:
    - read-only vs writable address behavior
    - node-local pending visibility
    - `eth_getTransactionCount` compatibility limitations
    - Canopy replay-protection / nonce model

  ## Why

  CEXs and Ethereum wallets assume standard Ethereum behavior in a few critical areas:
  - tx hash returned from broadcast matches the signed raw tx
  - receipts are mined-or-`null`
  - tx and receipt objects have Ethereum-compatible shapes
  - `eth_getTransactionCount` returns a usable next nonce
  - pending and mined lookup paths behave predictably

  Before this change, those assumptions were violated in several places, which created onboarding friction and false wallet behaviors like dropped/replaced interpretations.

  ## Validation

  Tested with:
  - `go test ./store ./cmd/rpc ./fsm`

  This PR was also exercised against local MetaMask/CEX-style flows during development of the nonce and tx lifecycle fixes.

### DIFF
--- a/cmd/rpc/eth.go
+++ b/cmd/rpc/eth.go
@@ -57,6 +57,8 @@ func (s *Server) EthereumHandler(w http.ResponseWriter, r *http.Request, _ httpr
 		ethResponse, err = s.NetListening(args)
 	case `net_peerCount`:
 		ethResponse, err = s.NetPeerCount(args)
+	case `eth_protocolVersion`:
+		ethResponse, err = s.EthProtocolVersion(args)
 	case `eth_syncing`:
 		ethResponse, err = s.EthSyncing(args)
 	case `eth_chainId`:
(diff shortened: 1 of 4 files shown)


andrewnguyen22:
Also tested with:
- https://github.com/canopy-network/ethrpc-test-harness

rem1niscence (approved):

PR #412, opened 2026-06-14
Audit fixes

andrewnguyen22:
## Summary

  - Tighten SafeNode() payload matching so either hash mismatch rejects the proposal.
  - Preserve queued inbox messages on overflow by dropping only the newest message.
  - Serialize plugin outbound IPC writes on the shared connection.
  - Use restrictive file modes for private keys, keystores, config, and shared JSON artifacts.
  - Keep store version state aligned with durable commit success/failure.
  - Add and adjust tests around pacemaker signatures, HighQC validation, and SafeNode behavior.

  ## Not Included

  - No change to the non-signer settlement behavior.
  - No change to plugin write-timeout / fail-stop behavior.
  - No change to proposer Timestamp / RcBuildHeight signature coverage.

  ## Validation

  go test ./controller ./bft ./p2p ./lib ./lib/crypto ./store

  ## Notes

  - The pacemaker signature change is a message-format compatibility change and should be merged and deployed coherently rather than mixed across versions.

### DIFF
--- a/bft/msg.go
+++ b/bft/msg.go
@@ -98,6 +98,23 @@ func (b *BFT) CheckProposerMessage(x *Message, p *validateMessageParams) (isPart
 	if err != nil {
 		return
 	}
+	if x.HighQc != nil {
+		if err = x.HighQc.CheckBasic(); err != nil {
+			return false, err
+		}
+		highQCVals, highQCView := p.vals, p.view
+		if x.HighQc.Header.RootHeight != p.rootHeight {
+			b.Controller.Lock()
+			highQCVals, err = b.LoadCommittee(b.LoadRootChainId(x.HighQc.Header.Height), x.HighQc.Header.RootHeight)
+			b.Controller.Unlock()
+			if err != nil {
+				return false, err
+			}
+		}
+		if err = x.HighQc.CheckHighQC(lib.GlobalMaxBlockSize, highQCView, p.rootHeightUpdated, highQCVals); err != nil {
+			return false, err
+		}
+	}
 	// A QC from a different root height must not justify a proposer message for the local view.
 	// Still allow storing it as partial-QC evidence (see AddPartialQC) if it's not +2/3 majority.
 	if x.Qc.Header.RootHeight != p.rootHeight {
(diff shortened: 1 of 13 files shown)


bft/msg.go
```diff
@@ -301,25 +318,27 @@ func (b *BFT) GetValidateMessageParams(msg *Message) (*validateMessageParams, li
...
-		vals:           b.ValidatorSet,
-		rootHeight:     b.RootHeight,
-		height:         b.Height,
-		cHeightUpdated: b.CommitteeData.LastChainHeightUpdated,
-		blockHash:      blockHash,
-		resultsHash:    resultsHash,
+		view:              b.View.Copy(),
+		vals:              b.ValidatorSet,
+		rootHeight:        b.RootHeight,
+		rootHeightUpdated: b.CommitteeData.LastRootHeightUpdated,
```
rem1niscence, line 324: what if a root chain and a nested chain are syncing and the nested chain, during sync, passes the root chain, wouldn't this break the sync as the `LastRootHeightUpdated` would be higher than the expected one?

store/store.go
```diff
@@ -249,29 +249,34 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
...
 	// finally commit the entire Transaction to the actual DB under the proper version (height) number
 	if err := s.db.Apply(s.writer, pebble.NoSync); err != nil {
-		return nil, ErrCommitDB(err)
+		commitErr := ErrCommitDB(err)
+		s.Reset()
+		return nil, commitErr
 	}
 	// update the metrics once complete
 	s.metrics.UpdateStoreMetrics(int64(size), int64(count), time.Time{}, startTime)
+	s.version = nextVersion
```
rem1niscence, line 279: `s.Reset()` also gets the current `s.version` and increases by one, this would result in 2 version increases per commit

rem1niscence (changes requested):

store/store.go
andrewnguyen22: `s.version` only advances once in `Commit()` 

`Reset()` is only preparing the next transaction writer with `writeVersion = s.version + 1`

bft/msg.go
andrewnguyen22: I think the sync flag already makes the sync case impossible

rem1niscence (approved):

PR #414, opened 2026-06-16
Custom rpc endpoints plugin

pablocampogo:
## Summary

Adds a detached, read-only query path from plugins to Canopy so plugin builders can serve their own custom RPC endpoints for chain-specific data, without coupling Canopy core to plugin-specific routes.

## What's added

**Core**
- New `PluginQueryRequest`/`PluginQueryResponse` proto messages + wire fields (`lib/.proto/plugin.proto`).
- `PluginQueryProvider` + `handleQueryRequest` in `lib/plugin.go`, backed by a controller-owned `TimeMachine` snapshot adapter (`controller/controller.go`). Detached queries are read-only (no `StateWrite`).
- Prefix-collision guard (two layers): plugins declare `custom_state_prefixes`; core **panics at handshake** if any collides with a core-reserved prefix (1–15), and `FSM.StateWrite` rejects writes to reserved prefixes (`fsm/state.go`, `fsm/plugin_guard_test.go`).

**Plugin SDKs (go, python, typescript, kotlin, csharp)**
- Detached `QueryState(height, read)` client method + transport.
- `rpcAddress` config field and a **skeleton HTTP server** (`rpc.*`) that starts with no routes, ready for builders to add their own endpoints.
- `custom_state_prefixes` config plumbing.

**Docs & tutorial**
- `TUTORIAL.md` / `AGENTS.md` for every plugin document how to add custom RPC endpoints, including a full worked faucet/reward example that lives only in each plugin's `tutorial/` folder (with integration tests).


lib/plugin.go
```diff
@@ -352,6 +407,57 @@ func (p *Plugin) handleStateReadRequest(msg *PluginToFSM) ErrorI {
...
+	p.l.Unlock()
+	// build the response container
+	response := new(PluginQueryResponse)
+	// check if a query provider is registered
+	if provider == nil {
+		p.log.Debugf("handleQueryRequest() no query provider registered for message ID %d", msg.Id)
+		response.Error = NewPluginError(ErrNoPluginQueryProvider())
+	} else {
+		// execute the detached read-only query
+		read, err := provider.QueryState(request.GetHeight(), request.GetRead())
```
andrewnguyen22, line 441: This forwards `request.GetRead()` straight into the quer provider, and `state.go line 680` dereferences the request immediately.

An empty `PluginQueryRequest` will panic the node instead of returning a plugin error - should be nil-guarded

controller/controller.go
```diff
@@ -327,10 +327,32 @@ func (c *Controller) PluginConnectSync() {
...
+// with Canopy's historical read-only snapshots (TimeMachine). It is the live-node-owned adapter
+// that lets plugin builders implement custom RPC endpoints without a tx/block in flight.
+type pluginQueryProvider struct {
+	controller *Controller
+}
+
+// QueryState() executes a read-only state read against a TimeMachine snapshot at the given height (0 = latest committed)
+func (p *pluginQueryProvider) QueryState(height uint64, request *lib.PluginStateReadRequest) (lib.PluginStateReadResponse, lib.ErrorI) {
+	// create a read-only state snapshot at the requested height
+	sm, err := p.controller.FSM.TimeMachine(height)
```
andrewnguyen22, line 350: This can panic on a fresh node before the first committed block (right after genesis). 

This discards whatever `TimeMachine(height)` returns, but `line 327` returns the live FSM itself when the chain is still at `H0`. So on a fresh node this can run defer `sm.Discard()` against *the live store*. 

The new SDK mains start the plugin RPC server immediately after handshake, a custom endpoint hit before the first commit can trigger this.

andrewnguyen22 (changes requested):

andrewnguyen22 (approved):

PR #419, opened 2026-06-25
Make JSON file writes atomic

### LINKED ISSUE #189: [TRACKING ONLY] Unable to unmarshal `book.json`
### Describe the bug
Normal operation then suddenly `unable to unmarshal peer book`

Image

Occurred on a non-validator full node on the canary network

### Steps to reproduce
Unknown if possible

### Expected behavior
Peer book flush and unmarshal is safe

### Screenshots
See above

### Environment
- **OS**: Linux + docker
- **Software Version**: alpha-0.0.1

### Additional context
Tracking only on this issue

#### Required Tags
- Priority: Low
- Module: Multi-Module

giwaov:
## Summary

- Add an atomic file-write helper that writes through a temporary file in the destination directory before renaming it into place.
- Use the helper for shared JSON persistence, config writes, and peer-book writes.
- Add focused coverage for replacement behavior and parent directory creation.

## Why

Fixes #189. A crash or interrupted write could leave persisted JSON files such as the peer book truncated, causing startup failures when the file was read again.

## Testing

- git diff --check

Go tests were not run locally because go is not installed on this machine.

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -360,8 +360,8 @@ func (c Config) WriteToFile(filepath string) error {
 		// exit with error
 		return err
 	}
-	// write the config.json file to the data directory
-	return os.WriteFile(filepath, jsonBytes, os.ModePerm)
+	// atomically write the config.json file to the data directory
+	return WriteFileAtomic(filepath, jsonBytes, os.ModePerm)
 }
 
 // NewConfigFromFile() populates a Config object from a JSON file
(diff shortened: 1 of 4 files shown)


giwaov:
Hi maintainers, could you please review this when convenient? It focuses on the truncated JSON persistence issue from #189 by using temp-file-and-rename writes for the shared JSON paths. I’m happy to tighten or split anything if you’d like a smaller scope.

PR #423, opened 2026-06-25
fix(p2p): replace unsynchronized dialing counter with atomic int32

amathxbt:
## Bug

In `p2p/p2p.go`, `DialForOutboundPeers()` tracks in-flight dial attempts with a plain `int`:

```go
dialing := 0
// ...
go func() {
    dialing++  // ← written from goroutine, read from main loop — data race
    p.DialWithBackoff(peerAddress, true)
}()
```

The counter is incremented inside spawned goroutines while the main loop reads it concurrently — with no mutex or atomic protection. This is a data race confirmed by `go test -race`.

**Result:** the node can significantly exceed `MaxOutbound` peers, opening unnecessary connections, wasting bandwidth, and destabilising the network layer.

## Fix

Replace `dialing int` with `var dialing int32` and use `atomic.AddInt32` / `atomic.LoadInt32` at every read and write site.

## Impact

Critical — data race in the P2P layer causes unbounded outbound connections.

### DIFF
--- a/p2p/p2p.go
+++ b/p2p/p2p.go
@@ -211,6 +211,9 @@ func (p *P2P) DialForOutboundPeers() {
 			atomic.AddInt32(&dialing, 1)
 			// dial the peer with exponential backoff
 			p.DialWithBackoff(peerAddress, true)
+			// decrement after DialWithBackoff completes so the outbound capacity
+			// loop does not permanently treat this dial as still in-flight.
+			atomic.AddInt32(&dialing, -1)
 		}()
 	}
 	// Continuous service until program exit, dial timeout loop frequency for resource break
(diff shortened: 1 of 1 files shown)


andrewnguyen22 (changes requested):
The finding is real **dialing is accessed from multiple goroutines in p2p.go**. 

This is directionally correct - but the PR is incomplete:

  1. It still reads p.PeerSet.outbound directly without the PeerSet lock at `p2p/p2p.go:221`, while outbound is written under peer-set mutation paths. That leaves a data race.

  2. The initial configured DialPeers goroutines increment dialing but never decrement it after DialWithBackoff returns. The PR preserves that bug, just with atomics. That can make the later outbound loop permanently think dials are still in flight.

amathxbt:
Thanks for the thorough review @andrewnguyen22 — both points are valid. Updated the branch with:

1. **Missing `dialing` decrement** — added `atomic.AddInt32(&dialing, -1)` at the end of each initial DialPeers goroutine, mirroring the `defer` decrement already used in the outbound capacity loop. Without this the counter stays permanently elevated after the configured peers connect, blocking all future outbound dials.

2. **`p.PeerSet.outbound` read without lock** — wrapped the read at line 222 in `p.PeerSet.mux.RLock()` / `RUnlock()`, consistent with how outbound is guarded elsewhere in the peer-set paths.

PR #426, opened 2026-06-25
fix(fsm): guard against empty PaymentPercents slice in HandleCertificateResults

amathxbt:
## Bug

In `fsm/automatic.go`, `HandleCertificateResults()` checks that `RewardRecipients` is non-nil but does not guard against `PaymentPercents` being an empty slice:

```go
if qc.Results.RewardRecipients == nil {
    return lib.ErrNilRewardRecipients()
}
// No check that PaymentPercents is non-empty ↓
for i, p := range results.RewardRecipients.PaymentPercents {
    results.RewardRecipients.PaymentPercents[i].Percent = ...
}
```

An empty `PaymentPercents` slice causes the reward-distribution loop to silently skip all payments — validators complete work but receive **no rewards**, and the block is accepted without error. A malicious or buggy proposer can craft a `CertificateResult` with a non-nil `RewardRecipients` but zero `PaymentPercents` to silently starve the committee of rewards for that block height.

## Fix

Return `ErrInvalidPercentAllocation` when `PaymentPercents` is empty after the existing nil guard, making the invariant explicit and ensuring the block is rejected rather than silently misprocessed.

## Impact

Critical — malicious proposer can craft blocks that pass validation but deliver zero rewards to the committee.

### DIFF
--- a/fsm/automatic.go
+++ b/fsm/automatic.go
@@ -144,8 +144,11 @@ func (s *StateMachine) HandleCertificateResults(qc *lib.QuorumCertificate, commi
 	if qc.Results.RewardRecipients == nil {
 		return lib.ErrNilRewardRecipients()
 	}
-	// guard against an empty PaymentPercents slice causing a silent no-op reward distribution
-	if len(qc.Results.RewardRecipients.PaymentPercents) == 0 && qc.Results.RewardRecipients != nil {
+	// Guard: an empty PaymentPercents slice would pass the nil check above but be
+	// silently persisted by UpsertCommitteeData, skipped by DistributeCommitteeRewards,
+	// and distort later reward accounting (NumberOfSamples accumulates without recipients).
+	// RewardRecipients non-nil is already guaranteed above; no need to re-check here.
+	if len(qc.Results.RewardRecipients.PaymentPercents) == 0 {
 		return lib.ErrInvalidPercentAllocation()
 	}
 	// ensure the committee isn't retired
--- a/fsm/automatic.go
+++ b/fsm/automatic.go
@@ -144,6 +144,10 @@ func (s *StateMachine) HandleCertificateResults(qc *lib.QuorumCertificate, commi
 	if qc.Results.RewardRecipients == nil {
 		return lib.ErrNilRewardRecipients()
 	}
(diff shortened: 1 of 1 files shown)


andrewnguyen22 (changes requested):
This part is real:
  - `HandleCertificateResults()` only checks `RewardRecipients != nil` at `fsm/automatic.go:152`. An empty PaymentPercents slice can pass into
  `UpsertCommitteeData()` at `fsm/automatic.go:224`.

  **But this does not “silently burn all block rewards on every block” immediately.** `DistributeCommitteeRewards()` explicitly skips empty PaymentPercents at `fsm/committee.go:142`. 

The more accurate risk is malformed/empty reward samples can be persisted, which can distort later reward accounting or cause later under-distribution/burn when non-empty samples are eventually distributed.

fsm/automatic.go
```diff
@@ -144,6 +144,10 @@ func (s *StateMachine) HandleCertificateResults(qc *lib.QuorumCertificate, commi
 	if qc.Results.RewardRecipients == nil {
 		return lib.ErrNilRewardRecipients()
 	}
+	// guard against an empty PaymentPercents slice causing a silent no-op reward distribution
+	if len(qc.Results.RewardRecipients.PaymentPercents) == 0 && qc.Results.RewardRecipients != nil {
```
andrewnguyen22: Use the existing shared validation instead of an ad hoc check:
```
  if err := qc.Results.RewardRecipients.CheckBasic(); err != nil {
      return err
  }
```

amathxbt:
Thanks @andrewnguyen22 you're right that the immediate-burn description was overstated. Updated the branch with:

- Removed the redundant `&& qc.Results.RewardRecipients != nil` from the guard (RewardRecipients non-nil is already confirmed two lines above, so the second condition was dead code).
- Rewrote the comment to accurately describe the actual risk: an empty `PaymentPercents` slice passes the nil guard, gets persisted by `UpsertCommitteeData`, is silently skipped by `DistributeCommitteeRewards` (line 142 of committee.go), and causes `NumberOfSamples` to accumulate without corresponding recipients — distorting later reward accounting rather than causing an immediate burn.

PR #437, opened 2026-06-28
Handle empty validator set in monitoring

### LINKED ISSUE #431: wallet: /monitoring page surfaces 'RPC 400' fatal error on validator-set returning code 29 (no validators in the set)
## Summary
`/monitoring` page renders a generic **"Monitoring Error RPC 400"** when `/v1/query/validator-set` returns `code: 29 — there are no validators in the set`. For chains with no staked committee validators (e.g. single-validator deployments) this is the normal/expected response, not an error worth surfacing as a fatal monitoring failure.

## Repro
- Single-validator chain, healthy + producing blocks
- Open `/monitoring` page in the canopy wallet UI

## Observed
- Page renders a red box: **"Monitoring Error — RPC 400"**
- Network panel shows `/v1/query/validator-set` returning HTTP 400 with:
  ```json
  {
    "code": 29,
    "module": "main",
    "msg": "there are no validators in the set"
  }
  ```
- Other RPC calls succeed (height, params, consensus-info, peer-info, resource-usage, log all 200)

## Expected
The monitoring page should:
1. Detect `code: 29` (`ErrNoValidators`) specifically
2. Render gracefully (e.g. "No committee validators staked yet" or "Single-validator chain — committee empty")
3. Continue showing the rest of the monitoring data that DID load

## Root cause
- Server: `lib/error.go` → `ErrNoValidators()` (HTTP 400, code 29) — this is the canonical response, expected when the committee has 0 staked validators
- Wallet: `cmd/rpc/web/wallet/public/plugin/canopy/chain.json` defines `validatorSet` query but doesn't appear to special-case the empty/zero-validators response

## Suggested fix
In the wallet's monitoring page renderer, treat `code: 29` from `/v1/query/validator-set` as a non-fatal state and show an empty/zero-validators panel instead of the generic error box.

## Reproduction details
- Chain: universal-databases-98799, chain_id 243
- Pod image: `canonical-v6` (rebuilt today from RichardJamesLopez/canopy@e1d9ee47)
- Wallet host: `data-wallet.val-a.grad.dev.app.canopynetwork.org`
- Reproduced via curl with payloads `{}`, `{"height":0}`, `{"chainId":98799}`, `{"id":98799}`, `{"id":98799,"height":50000}` — all return the same 400/code:29

giwaov:
## Summary

- Preserve RPC error metadata in the wallet DS fetcher so callers can inspect server error codes.
- Treat validator-set error code 29 as an empty validator set in monitoring data loading.
- Keep other validator-set and monitoring fetch failures fatal.

## Why

Fixes #431. A chain with no staked committee validators returns the canonical no-validators response, but the monitoring page treated it as a fatal RPC 400 and hid the other data that loaded successfully.

## Testing

- npm run build
- git diff --check

### DIFF
--- a/cmd/rpc/web/wallet/src/core/dsCore.ts
+++ b/cmd/rpc/web/wallet/src/core/dsCore.ts
@@ -167,7 +167,27 @@ export async function fetchDsOnce<T=any>(chain: ChainLike, key: string, ctx?: Re
     const { url, init } = buildRequest(chain, leaf, ctx)
     if (!url) throw new Error(`Invalid DS url for key ${key}`)
     const res = await fetch(url, init)
-    if (!res.ok) throw new Error(`RPC ${res.status}`)
+    if (!res.ok) {
+        const ct = res.headers.get('content-type') || ''
+        let response: any
+        try {
+            response = ct.includes('application/json') ? await res.json() : await res.text()
+        } catch {
+            response = undefined
+        }
+
+        const message =
+            response && typeof response === 'object' && 'msg' in response
+                ? `RPC ${res.status}: ${response.msg}`
+                : `RPC ${res.status}`
+
+        throw Object.assign(new Error(message), {
+            status: res.status,
+            code: response && typeof response === 'object' ? response.code : undefined,
+            module: response && typeof response === 'object' ? response.module : undefined,
+            response,
+        })
+    }
     const parsed = await parseResponse(res, leaf)
     return parsed as T
 }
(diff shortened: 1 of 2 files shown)


giwaov:
Hi maintainers, could you please take a look when you have a chance? This one addresses the reported monitoring-page failure for empty validator sets and keeps the rest of the monitoring data visible. I’m happy to adjust the scope if you’d prefer a different handling for code 29.

PR #457, opened 2026-07-04
fix(p2p): make P2P gossip mode flag atomic to eliminate a data race

amathxbt:
## Bug

`P2P.gossip` (in `p2p/p2p.go`) is a plain `bool` field read and written from different goroutines without synchronization:

```go
type P2P struct {
	...
	gossip bool // whether gossip mode is active
	...
}

func (p *P2P) SetGossipMode(gossip bool) {
	p.gossip = gossip
}

func (p *P2P) GossipMode() bool {
	return p.gossip
}
```

`SetGossipMode()` is called from `Controller` in `controller/consensus.go` (during must-connect/peer-count updates), while `GossipMode()` is read concurrently from multiple other places in `controller/consensus.go` (`ShouldGossip`, and the self-send/replica-broadcast decision in the block-proposal path) as part of normal BFT/consensus operation. These calls happen on different goroutines (consensus loop vs. peer-management callbacks) with no shared lock, which is a genuine unsynchronized concurrent read/write on a plain `bool` — undefined behavior per the Go memory model, and flagged by the race detector.

## Fix

Changed `gossip` from `bool` to `atomic.Bool` and updated `SetGossipMode`/`GossipMode` to use `Store`/`Load`. This is a minimal, drop-in fix: `atomic.Bool`'s zero value is `false`, matching the previous default, and both accessor methods keep their existing signatures so no call sites outside this file need to change.

## Testing

Confirmed `p.gossip` was only referenced within `p2p/p2p.go` via `SetGossipMode`/`GossipMode`, and that all external callers (in `controller/consensus.go`) only go through these two methods, so the internal type change is fully encapsulated.

### DIFF
--- a/p2p/p2p.go
+++ b/p2p/p2p.go
@@ -1,467 +1,468 @@
 package p2p
 
 import (
-	"bytes"
-	"fmt"
-	"math/rand"
-	"net"
-	"runtime/debug"
-	"slices"
-	"strings"
-	"sync"
-	"time"
+        "bytes"
+        "fmt"
+        "math/rand"
+        "net"
+        "runtime/debug"
+        "slices"
+        "strings"
+        "sync"
+        "sync/atomic"
+        "time"
 
-	"github.com/canopy-network/canopy/lib"
-	"github.com/canopy-network/canopy/lib/crypto"
-	"github.com/cenkalti/backoff/v4"
-	"github.com/phuslu/iploc"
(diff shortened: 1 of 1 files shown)


rem1niscence:
Could you please fix the merge conflict and fix the indentation? the rationale is good but the code is changing the indentation of all the changed files which is not desired

PR #458, opened 2026-07-04
Parse RPC listen host safely

giwaov:
## Summary

- Parse the configured listen host with net.SplitHostPort before starting RPC servers.
- Reuse the parsed host for wallet and explorer static server logging.
- Add coverage for IPv4, wildcard, IPv6, and host-only listen addresses.

## Why

The previous strings.Split call split on every colon. IPv6 listen addresses such as [::]:9001 were reduced to an invalid host fragment instead of the configured host.

## Testing

- git diff --check
- Not run: Go toolchain is not installed in this environment.


cmd/rpc/server_test.go
```diff
@@ -0,0 +1,24 @@
+package rpc
```
rem1niscence: Please remove this test, the new `listenHost` function uses two standard library methods to work, these are guaranteed to work as expected and no test is needed to confirm their behavior. Once you do remove this I'll be happy to merge

rem1niscence (changes requested):

PR #470, opened 2026-07-13
Docker betterments

pablocampogo:
Add multi arch support
Change plugin images to be their own image instead of a tag


.docker/auto-update/Dockerfile
```diff
@@ -6,6 +6,7 @@ ARG BUILD_PATH=cmd/cli
 ARG GO_BIN_DIR
 ARG BUILD_LOCAL=False
 ARG BIN_PATH=/bin/cli
+ARG TARGETARCH
```
rem1niscence: This can be removed, go uses the arch of the system that is building it so when you tell docker to build in `amd64` it uses `amd64`, same for `arm64`. This is not used on the building CI either because of the aforementioned reason, otherwise this is just dangling code

rem1niscence (approved):

PR #472, opened 2026-07-15
Fix/rlp v2 audit followups

andrewnguyen22:
## What this does

This adds RLP.V2 transactions with account-based nonce protection for Ethereum wallets.

RLP.V2 stores the signed nonce directly on the Canopy transaction and advances a nonce floor on the sender’s account after execution. The nonce does not need to be consecutive: committing nonce `N` invalidates that nonce and everything below it.

Legacy RLP and RLP.V2 use different Ethereum chain ID domains, so a signature for one cannot be reused as the other. Legacy decoding remains available for replaying historical blocks, while new legacy transactions are disabled at protocol version 2.

  ## Ethereum RPC changes

  - `eth_sendRawTransaction` now creates RLP.V2 transactions.
  - `eth_getTransactionCount` reads the committed account nonce and includes local pending transactions for `"pending"`.
  - Added `eth_feeHistory` and `eth_maxPriorityFeePerGas` for wallet compatibility.
  - Added same-nonce replacement support. Both the fee cap and tip cap must increase by at least 10%.
  - Replaced transactions are removed only after the replacement survives mempool insertion.
  - Stale entries in the short-lived RPC cache no longer block replacements.
  - The pending transaction cache is bounded and expires entries after two minutes.

  ## Follow-up fixes

  This also includes the issues found during review:

  - Drained accounts are deleted when their balance and nonce are both zero. RLP.V2 accounts with a committed nonce are retained.
  - Normal fee-based mempool eviction no longer goes through invalid-transaction handling.
  - Vesting math now avoids `uint64` multiplication overflow.
  - Equal-fee mempool transactions keep their arrival order.
  - Plugin implementations preserve the new account nonce.
  - The transaction nonce uses protobuf field `10`; plugin-specific transaction fields are safe from field `25` onward.

  The Go, Python, TypeScript, Kotlin, and C# plugin schemas were updated together.

  ## Rollout

Validators and configured plugins need to be upgraded together before RLP.V2 submissions begin. No account migration or nonce backfill is required.

RLP.V2 submission or replay during the transition window is an accepted rollout assumption.


store/indexer.go
```diff
@@ -307,7 +302,7 @@ func indexedTxHashes(result *lib.TxResult) ([][]byte, lib.ErrorI) {
 
 // ethTxHash() returns the canonical Ethereum tx hash for an RLP-backed transaction.
 func ethTxHash(tx *lib.Transaction) []byte {
-	if tx == nil || tx.Memo != "RLP" || tx.Signature == nil || len(tx.Signature.Signature) == 0 {
+	if tx == nil || (tx.Memo != "RLP" && tx.Memo != "RLP.V2") || tx.Signature == nil || len(tx.Signature.Signature) == 0 {
```
pablocampogo: Don't use hardcoded memo here. Use the fsm.RLPIndicator / fsm.RLPV2Indicator / fsm.IsRLPMemo functions

cmd/rpc/eth.go
```diff
@@ -274,7 +289,7 @@ func (s *Server) EthGetBalance(args []any) (result any, err error) {
 	// create a read-only state for the block tag
 	err = s.readOnlyState(height, func(state *fsm.StateMachine) (e lib.ErrorI) {
 		// get the balance for the address
-		balance, e := state.GetAccountSpendableBalance(address)
+		balance, e := state.GetAccountBalance(address)
```
pablocampogo, line 292: Shouldn't this stay with the actual spendable value?

pablocampogo (changes requested):

store/indexer.go
andrewnguyen22: Fixed

cmd/rpc/eth.go
andrewnguyen22: No this is the behavior we want

rem1niscence:
tests failing with: 

> imports github.com/canopy-network/canopy/fsm from indexer.go: import cycle not allowed in test

andrewnguyen22:
> tests failing with:
> 
> > imports github.com/canopy-network/canopy/fsm from indexer.go: import cycle not allowed in test

Fixed @rem1niscence

pablocampogo (approved):

controller/tx.go
```diff
@@ -410,10 +425,41 @@ func (c *Controller) GetPendingTxByHash(hash string) (*lib.TxResult, bool) {
...
+}
+
+// GetPendingNonce recommends the next nonce after the latest validated mempool snapshot.
+func (c *Controller) GetPendingNonce(address crypto.AddressI, confirmed uint64) uint64 {
+	if c == nil || c.Mempool == nil || c.Mempool.L == nil || !c.Mempool.L.TryLock() {
+		return confirmed
+	}
+	defer c.Mempool.L.Unlock()
+	for _, result := range c.Mempool.cachedResults {
+		if result == nil || result.Transaction == nil || result.Transaction.Memo != fsm.RLPV2Indicator ||
```
rem1niscence, line 454: As the pending nonce is somewhat dependant on the mempool cache, is it expected that if an user is connected to a canopy node (A) of X chain to process eth txs and for whatever reason goes to a different node (B) to perform txs, if it returns back to the first node (A) those transactions will fail due to a lower nonce err?

fsm/state.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	CurrentProtocolVersion         = 1
+	CurrentProtocolVersion         = 2
```
rem1niscence, line 17: safe to assume that a protocol bump must be done right after merging this?

rem1niscence (approved):
Although most of the review was focused on understanding the PR itself and styling issues as I lack the full implementation details to correct on logic. This seems pretty complete and throroughly tested. Just left a few questions to confirm behavior

controller/tx.go
andrewnguyen22: Yes, this is expected and how ethereum behaves

fsm/state.go
andrewnguyen22: A protocol bump doesn't need to be done immediately - this is in preparation if we do submit the protocol change transaction

PR #478, opened 2026-07-21
feat(dex): cap LP providers with deterministic replacement

andrewnguyen22:
## Summary

- cap liquidity pools at 5,000 providers and deterministically replace the lowest-point provider when a better-funded newcomer arrives
- migrate legacy pools of up to 50,000 providers on the first capped deposit while enforcing the 5,000-provider certificate limit for every path, including liveness fallback
- aggregate newcomer deposits by address, use receipt-hash-seeded tie breaking, give evictions distinct deterministic IDs, and expose the auto-withdrawal percentage to indexers
- select and apply replacements in memory and persist the pool once, avoiding thousands of complete 5,000-provider pool rewrites
- permit certificate-result transactions above the individual transaction limit while retaining the collective block/mempool byte bound

## Validation

- all non-audit packages: `go list -e ./... | rg -v '/audit/' | xargs go test`
- `go test -race ./fsm -run 'TestHandleBatch|Test.*Capped|Test.*Liquidity|TestHandleDexBatchOrdersEnforcesSettlementCap' -count=1`
- `go test -race ./lib -run 'TestCertificateResultsMayExceedIndividualMaxTxSize|TestDexBatch' -count=1`
- two-node Docker integration: normal cross-chain DEX trade completed successfully
- legacy migration: both 50,000-provider pools converged to the same 5,000-provider state
- maximum newcomer batch on a 1 CPU / 2 GiB node: 1.67s proposal construction and 1.72s validation
- indexing check: 4,999 accepted deposits and 4,999 unique 100% auto-withdrawals, with unique order IDs and no zero-point providers

The repository-wide `go test ./...` also passes all normal packages; its only failures are checked-in `audit/**/artifacts` repro directories that are not standalone buildable packages.


lib/certificate.go
```diff
@@ -913,7 +913,7 @@ func (x *DexBatch) CheckBasic() (err ErrorI) {
 	if len(x.Receipts) > MaxReceipts {
 		return ErrTooManyDexReceipts()
 	}
-	// ensure there's not too many receipts
+	// ensure there's not too many liquidity providers
 	if len(x.PoolPoints) > MaxLiquidityProviders {
 		return ErrTooManyDexReceipts()
```
pablocampogo: The error here should be ErrTooManyLiquidityProviders

fsm/dex.go
```diff
@@ -632,7 +671,163 @@ func (s *StateMachine) HandleBatchDeposit(batch *lib.DexBatch, chainId uint64, x
...
+
+// handleCappedBatchDeposit() handles the 5k limit
+func (s *StateMachine) handleCappedBatchDeposit(batch *lib.DexBatch, p *Pool, chainId uint64, x, y *uint64, local bool) (bool, lib.ErrorI) {
+	var err lib.ErrorI
+	if excess := len(p.Points) - lib.MaxLiquidityProviders; excess > 0 {
+		sort.Slice(p.Points, func(i, j int) bool {
+			if p.Points[i].Points != p.Points[j].Points {
+				return p.Points[i].Points < p.Points[j].Points
+			}
+			return bytes.Compare(crypto.Hash(append(bytes.Clone(batch.ReceiptHash), p.Points[i].Address...)), crypto.Hash(append(bytes.Clone(batch.ReceiptHash), p.Points[j].Address...))) < 0
```
pablocampogo: `crypto.Hash(append(bytes.Clone(batch.ReceiptHash), p.Points[i].Address...))` this is used in 4 places in this file. I think it's worth separating it to a helper function that also explains what this actually does

fsm/dex.go
```diff
@@ -632,7 +671,163 @@ func (s *StateMachine) HandleBatchDeposit(batch *lib.DexBatch, chainId uint64, x
...
 	// update the pool
-	return s.SetPool(p)
+	if persist {
+		return s.SetPool(p)
+	}
+	return nil
+}
+
+// handleCappedBatchDeposit() handles the 5k limit
+func (s *StateMachine) handleCappedBatchDeposit(batch *lib.DexBatch, p *Pool, chainId uint64, x, y *uint64, local bool) (bool, lib.ErrorI) {
```
pablocampogo, line 678: Add more in line comments to this function to keep the style of the rest of the codebase

fsm/dex.go
```diff
@@ -632,7 +671,163 @@ func (s *StateMachine) HandleBatchDeposit(batch *lib.DexBatch, chainId uint64, x
...
+	})
+	// create an apply deposit callback
+	apply := func(newcomer *candidate) lib.ErrorI {
+		return s.handleBatchDeposit(&lib.DexBatch{Deposits: newcomer.deposits}, chainId, x, y, local, false, p, false)
+	}
+	// for each newcomer
+	var lowest *lib.PoolPoints
+	for _, newcomer := range newcomers {
+		// if there's a free slot
+		if len(p.Points) < lib.MaxLiquidityProviders {
```
pablocampogo, line 741: This includes `deadAddr` making the real limit 4,999. Is this intended?

fsm/dex.go
```diff
@@ -632,7 +671,163 @@ func (s *StateMachine) HandleBatchDeposit(batch *lib.DexBatch, chainId uint64, x
...
+	// determine incumbents vs newcomers
+	var incumbents []*lib.DexLiquidityDeposit
+	var newcomers []*candidate
+	byAddress := make(map[string]*candidate)
+	for _, deposit := range batch.Deposits {
+		address := string(deposit.Address)
+		if providers[address] {
+			incumbents = append(incumbents, deposit)
+		} else {
+			newcomer := byAddress[address]
```
pablocampogo, line 705: Here for consistency/defensive programming we check if the amount is greater than 0 before adding a new comer

fsm/dex.go
```diff
@@ -632,7 +671,163 @@ func (s *StateMachine) HandleBatchDeposit(batch *lib.DexBatch, chainId uint64, x
...
+		xOut := lib.SafeMulDiv(*x, lowest.Points, p.TotalPoolPoints)
+		yOut := lib.SafeMulDiv(*y, lowest.Points, p.TotalPoolPoints)
+		// calculate the pool’s geometric-mean liquidity immediately before and after the newcomer’s
+		// hypothetical deposit, assuming the lowest LP has already been removed
+		oldK, newK := lib.SqrtProductUint64(*x-xOut, *y-yOut), lib.SqrtProductUint64(*x-xOut+newcomer.amount, *y-yOut)
+		if oldK == 0 {
+			return true, ErrInvalidLiquidityPool()
+		}
+		// calculate the newcomer's share assuming the lowest LP has already been removed
+		totalShare := lib.SafeMulDiv(p.TotalPoolPoints-lowest.GetPoints(), newK-oldK, oldK)
```
pablocampogo: This math or very similar math is being done in 2 places with kind of repeated code. These are here in the gate, at lines 819-823 in the eviction. This math should be moved to a shared function to prevent drifting variations in any of them in the future.

fsm/dex.go
```diff
@@ -472,10 +489,15 @@ func (s *StateMachine) HandleBatchWithdraw(batch *lib.DexBatch, counterChainId u
...
 		// emit withdraw event
-		if err = s.EventDexLiquidityWithdraw(w.Address, w.OrderId, payout, counter, points, counterChainId); err != nil {
+		if err = s.EventDexLiquidityWithdraw(w.Address, w.OrderId, payout, counter, points, w.Percent, counterChainId); err != nil {
 			return err
 		}
 	}
+	// second call removes providers whose points became zero during the current withdrawal
+	p.Points = slices.DeleteFunc(p.Points, func(point *lib.PoolPoints) bool { return point.Points == 0 })
+	if p.TotalPoolPoints == 0 {
+		return lib.ErrZeroLiquidityPool()
```
pablocampogo, line 507: Add a comment saying this should be unreachable because `deadAddr` holds points

fsm/dex.go
```diff
@@ -632,7 +671,163 @@ func (s *StateMachine) HandleBatchDeposit(batch *lib.DexBatch, chainId uint64, x
...
+					OrderId: crypto.ShortHash(lib.JoinLenPrefix([]byte("dex-lp-migration-v1"), batch.ReceiptHash, point.Address))})
+			}
+			if len(withdrawals) == excess {
+				break
+			}
+		}
+		if len(withdrawals) != excess {
+			return true, ErrInvalidLiquidityPool()
+		}
+		if err := s.HandleBatchWithdraw(&lib.DexBatch{Withdrawals: withdrawals}, chainId, x, y, local); err != nil {
```
pablocampogo: Here the private functionc `s.handleBatchWithdraw` with persisted=false should be used since this is causing multiple persists since at the end of this function the pool is persisted.

pablocampogo (changes requested):

fsm/dex.go
andrewnguyen22: Yes

fsm/dex.go
andrewnguyen22: Took another look and reverted this. It only saves one write during a one-time migration, so the extra branching wasn’t worth it.

fsm/dex.go
andrewnguyen22: Fixed in f59e0bd5. Extracted liquidityDepositPoints and now use the same geometric-mean minting calculation for both normal deposits and hypothetical post-eviction admission.

lib/certificate.go
andrewnguyen22: added `ErrTooManyLiquidityProviders` with its own error code and updated CheckBasic plus its test to assert that specific error

fsm/dex.go
andrewnguyen22: Added the unreachable invariant comment

fsm/dex.go
andrewnguyen22: Added inline comments

fsm/dex.go
andrewnguyen22: Fixed. Zero-amount deposits now follow the ordinary deposit path and are not classified or counted as newcomers

fsm/dex.go
andrewnguyen22: extracted `liquidityProviderTieBreakHash`

pablocampogo (approved):

fsm/dex.go
```diff
@@ -398,36 +406,53 @@ func (s *StateMachine) HandleDexBatchOrders(remoteBatch *lib.DexBatch, x, y *uin
...
-		return err
+	var err lib.ErrorI
+	if p == nil {
+		p, err = s.GetPool(counterChainId + LiquidityPoolAddend)
+		if err != nil {
+			return err
+		}
+	}
+	// removes legacy zero-point “ghost” LPs
+	p.Points = slices.DeleteFunc(p.Points, func(point *lib.PoolPoints) bool { return point.Points == 0 })
```
rem1niscence, line 427: why do this deletion at the beginning of the function when it will be called again after the withdrawals are processed? shouldn't the second call also clean the "ghost" LPs?

cmd/rpc/query.go
```diff
@@ -37,11 +37,15 @@ func (s *Server) Transaction(w http.ResponseWriter, r *http.Request, _ httproute
 // Transactions handles multiple transactions in a single request
 func (s *Server) Transactions(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	// create a slice to hold the incoming transactions
-	var txs []lib.TransactionI
+	var transactions []*lib.Transaction
```
rem1niscence, line 40: what's the logic behind using the concrete type? when I first made this change I used the interface in order to avoid the iteration done now to cast to the concrete type as a small optimization, any reason to do this?

rem1niscence (approved):
Please check the comments left

fsm/dex.go
andrewnguyen22: Yep, the second pass cleans old ghosts too if we actually reach it. The catch is the early return above: if the batch only touches zero-point holders, totalPointsToRemove is 0 and we never get to that second pass. So the first one clears old ghosts, and the second clears LPs this withdrawal just took to zero.

cmd/rpc/query.go
andrewnguyen22: Yeah, I wanted to keep the interface here too, but encoding/json can't instantiate entries in []lib.TransactionI, so normal tx objects fail to decode on the batch endpoint. Decoding into []*lib.Transaction gives it a concrete target. The loop is only wrapping pointers in the interface slice, not copying the txs, and keeps us from needing a custom decoder or a wider submitTxs change.

PR #499, opened 2026-08-04
Perf/indexer state change journal

andrewnguyen22:
## Summary

This changes the indexer-blobs account delta path so it no longer needs to scan the full account state at both the current and previous heights. When stateChangeJournalEnabled is turned on, each commit records the state keys it touched and the RPC uses the account keys from that journal to load only the relevant values. The setting is disabled by default.

Older heights without journal data still use the existing full-snapshot fallback. Height 2 also keeps the full snapshot so genesis accounts are preserved, and reward or slash event accounts are included even when their state did not change.

## Testing

- go test -count=1 ./lib ./store ./fsm ./cmd/rpc


store/store.go
```diff
@@ -262,6 +266,15 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
 	}
 	// collect LSS tombstones before Flush() clears the txn operations
 	lssDeleteKeys := s.collectLssDeleteKeys()
+	// Persist the keys touched by this commit outside consensus state. Indexer
+	// blob queries use this journal to avoid scanning millions of unchanged
+	// accounts at every height.
```
rem1niscence: the second part of the comment is unnecesary, it talks about a possible issue that might not be true for all chains, the comment should only address what the code is doing, not the why of somebody that benefits from it

cmd/rpc/query.go
```diff
@@ -623,6 +623,25 @@ func (s *Server) IndexerBlobsCached(height uint64) (*fsm.IndexerBlobs, []byte, l
 		return entry.deltaBlobs, entry.deltaBytes, nil
 	}
 
+	// Newer store versions persist the state keys touched by each commit. This
+	// makes account delta construction proportional to block activity instead
+	// of total account count. Heights committed before the journal was enabled
+	// transparently use the full-snapshot fallback below.
+	if journalDelta, available, journalErr := s.controller.FSM.IndexerBlobsFromStateChanges(height); journalErr != nil {
+		return nil, nil, journalErr
+	} else if available {
```
rem1niscence:
```suggestion
	journalDelta, available, journalErr := s.controller.FSM.IndexerBlobsFromStateChanges(height)
	if journalErr != nil {
		return nil, nil, journalErr
	}
	if available {
```
Nitpick to make it a bit easier to follow and go-idiomatic

store/store.go
```diff
@@ -292,6 +305,19 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
 	return
 }
 
+// recordStateChangeKeys snapshots the pending state transaction before Flush
+// clears it. Values are already available from the versioned state store, so
+// the journal only needs keys
+func (s *Store) recordStateChangeKeys(version uint64) lib.ErrorI {
```
rem1niscence, line 309: all the changed keys are saved, accounts, validators, smt nodes, and so on. Only the accounts are actually read from the journal, is it possible to add a filter to only save the accounts or at least exclude smt nodes which will likely never be read?

rem1niscence (changes requested):

store/store.go
andrewnguyen22: The journal does not currently save SMT nodes.

`recordStateChangeKeys()` iterates only over:

```text
s.ss.txn.ops
```

That contains application state such as accounts, validators, pools, parameters, DEX state, and plugin state. SMT nodes use `s.sc.store`, while blocks, transactions, and events use `s.Indexer.db`; neither is included.

Filtering to accounts is possible and would reduce write amplification and journal storage. However, the `store` package should not hard-code the FSM account prefix, as that would couple generic storage code to the application's key layout.

I say we keep the journal generic for now, especially if pools or validators may use it later.

Also, journaling is disabled by default, only operators who opt into faster indexer reads incur this additional storage cost

store/store.go
andrewnguyen22: Possibly, I thought it was important to justify why we have this exception - I'll remove it for now

rem1niscence (approved):

PR #505, opened 2026-08-07
feat(indexer): observability and robustness for the indexer-blobs RPC path

ezeike:
## Summary

Instrumentation and hardening for the `IndexerBlobs`/`IndexerBlob` RPC path, ahead of the
larger sparse-journal restructure in #506. No behavior change to the blob contents themselves.

- **RPC robustness**: cancel in-flight `IndexerBlobs` scans on client disconnect; raise the
  default RPC timeout from 3s to 30s to give large-scan requests room to complete.
- **`DeltaIndexerBlobs` perf**: replace the map-based diff with a sorted merge-walk.
- **Observability**: per-step timing (`ObserveIndexerBlobStep`) across time-machine setup,
  block fetch, account/validator iteration, and delta compute; path (`legacy`/`journal`) and
  tier (`hss`/`lss`) labels throughout; pebble LSM-tree metrics; previous-blob reuse hit/miss
  counters; cold-read visibility on the `indexer-blobs` endpoint.
- **Store fix**: give post-sync `CompactAll` its own compaction timeout, distinct from the
  regular compaction path.
- Misc: local-build k3s node deploy script.

## Stack

Base of a 3-PR stack: **#505 (this PR) → #506 → #507**. #506 builds the sparse
validators/non-signers journal restructure on top of this; #507 adds a `LoadCommittee`
memoization cache on top of that.


ezeike:
Closing — this draft's branch (`feat/indexer-blobs-observability`) was never updated past its original tier-label commit; the actual work described in this PR's body ended up stacked as separate PRs instead: #506 (journal restructure, validator/non-signer sparsity, totals cache) → #507 (committee-membership cache) → #508 (tier-label metric). Confirmed nothing references this PR (no commits, code, docs, or tracker entries), so closing rather than retargeting/updating it further.

lib/metrics.go
```diff
@@ -686,6 +795,57 @@ func NewMetricsServer(nodeAddress crypto.AddressI, chainID float64, softwareVers
 				Help: "Execution time of NewCertificateResults during CheckMempool proposal building",
 			}),
 		},
+		// INDEXER
+		IndexerMetrics: IndexerMetrics{
+			IndexerBlobColdReadTime: promauto.NewHistogram(prometheus.HistogramOpts{
+				Name: "canopy_indexer_blob_cold_read_time",
+				Help: "The time it takes to serve an indexer-blobs request that misses the cache (cold store read)",
+			}),
```
pablocampogo, line 777: This histogram doesn't set a `Buckets` field, so it falls back to `prometheus.DefBuckets`, whose largest finite bucket is 10s, but this PR raises the RPC timeout to 30s because these cold reads can take that long.
We already use custom buckets elsewhere in the code https://github.com/canopy-network/canopy/blob/e144b317f657e4ddcde2bc9e4943a11c157d9c32/lib/metrics.go#L359-L363 use this as reference to do the change

lib/metrics.go
```diff
@@ -686,6 +795,57 @@ func NewMetricsServer(nodeAddress crypto.AddressI, chainID float64, softwareVers
...
+				Help: "Total indexer-blobs requests whose 'previous' blob was reused from the prior request's cached current",
+			}),
+			IndexerBlobPreviousReuseMisses: promauto.NewCounter(prometheus.CounterOpts{
+				Name: "canopy_indexer_blob_previous_reuse_misses_total",
+				Help: "Total indexer-blobs requests whose 'previous' blob required its own fresh cold read",
+			}),
+			IndexerBlobStepTime: promauto.NewHistogramVec(prometheus.HistogramOpts{
+				Name: "canopy_indexer_blob_step_time",
+				Help: "The time each step of an indexer-blobs read takes, by step name, path (journal/legacy), and store tier (lss/hss/n_a)",
+			}, []string{"step", "path", "tier"}),
```
pablocampogo, line 806: This histogram doesn't set a `Buckets` field, so it falls back to `prometheus.DefBuckets`, whose largest finite bucket is 10s, but this PR raises the RPC timeout to 30s because these cold reads can take that long.
We already use custom buckets elsewhere in the code https://github.com/canopy-network/canopy/blob/e144b317f657e4ddcde2bc9e4943a11c157d9c32/lib/metrics.go#L359-L363 use this as reference to do the change

lib/metrics.go
```diff
@@ -686,6 +795,57 @@ func NewMetricsServer(nodeAddress crypto.AddressI, chainID float64, softwareVers
...
+			IndexerBlobPathTotal: promauto.NewCounterVec(prometheus.CounterOpts{
+				Name: "canopy_indexer_blob_path_total",
+				Help: "Total indexer-blobs requests by which path served them: cache_hit, journal, or legacy",
+			}, []string{"path"}),
+			IndexerBlobJournalBuildTime: promauto.NewHistogram(prometheus.HistogramOpts{
+				Name: "canopy_indexer_blob_journal_build_time",
+				Help: "The time it takes to serve an indexer-blobs request via the state-change journal (available=true)",
+			}),
```
pablocampogo, line 819: This histogram doesn't set a `Buckets` field, so it falls back to `prometheus.DefBuckets`, whose largest finite bucket is 10s, but this PR raises the RPC timeout to 30s because these cold reads can take that long.
We already use custom buckets elsewhere in the code https://github.com/canopy-network/canopy/blob/e144b317f657e4ddcde2bc9e4943a11c157d9c32/lib/metrics.go#L359-L363 use this as reference to do the change

lib/metrics.go
```diff
@@ -686,6 +795,57 @@ func NewMetricsServer(nodeAddress crypto.AddressI, chainID float64, softwareVers
...
+			IndexerBlobRequestTime: promauto.NewHistogramVec(prometheus.HistogramOpts{
+				Name: "canopy_indexer_blob_request_time",
+				Help: "The total time to serve an indexer-blobs request end-to-end, for every path (cache_hit/journal/legacy) -- compare directly against the client-observed RPC round-trip",
+			}, []string{"path"}),
```
pablocampogo, line 824: This histogram doesn't set a `Buckets` field, so it falls back to `prometheus.DefBuckets`, whose largest finite bucket is 10s, but this PR raises the RPC timeout to 30s because these cold reads can take that long.
We already use custom buckets elsewhere in the code https://github.com/canopy-network/canopy/blob/e144b317f657e4ddcde2bc9e4943a11c157d9c32/lib/metrics.go#L359-L363 use this as reference to do the change

cmd/rpc/query.go
```diff
@@ -643,8 +660,18 @@ func (s *Server) IndexerBlobsCached(height uint64) (*fsm.IndexerBlobs, []byte, l
...
+	// Cache miss and no journal available for this height: current/previous blobs
+	// must be built from the store, which for a height far behind the tip means a
+	// cold historical read. This is unbounded and not covered by the RPC
+	// write-deadline (the deadline keeps ticking while this call blocks, so a slow
+	// read here surfaces later as a "write tcp ...: i/o timeout" on the eventual
+	// w.Write, not here) -- log and record how long it actually took so a cold-path
+	// stall is visible after the fact both in logs and in canopy_indexer_blob_cold_read_time.
+	coldStart := time.Now()
+	defer s.controller.Metrics.RecordIndexerBlobCacheMiss(coldStart)
+	current, err := s.controller.FSM.IndexerBlob(ctx, height)
```
pablocampogo, line 688: If this function can return `ErrCancelled` the metrics `RecordIndexerBlobCancelled` would not get recorded. If that is not the case ignore the comment

cmd/rpc/query.go
```diff
@@ -643,8 +660,18 @@ func (s *Server) IndexerBlobsCached(height uint64) (*fsm.IndexerBlobs, []byte, l
...
+
+	// Cache miss and no journal available for this height: current/previous blobs
+	// must be built from the store, which for a height far behind the tip means a
+	// cold historical read. This is unbounded and not covered by the RPC
+	// write-deadline (the deadline keeps ticking while this call blocks, so a slow
+	// read here surfaces later as a "write tcp ...: i/o timeout" on the eventual
+	// w.Write, not here) -- log and record how long it actually took so a cold-path
+	// stall is visible after the fact both in logs and in canopy_indexer_blob_cold_read_time.
+	coldStart := time.Now()
+	defer s.controller.Metrics.RecordIndexerBlobCacheMiss(coldStart)
```
pablocampogo: This `cacheMiss` metrict even if the read was cancelled. Maybe we should add an if to prevent polluting the data

lib/config.go
```diff
@@ -152,7 +152,7 @@ func DefaultRPCConfig() RPCConfig {
 		ProfilingPort:              "6060",                     // the pprof profiling server is served on localhost:6060
 		RPCUrl:                     "http://localhost:50002",   // use a local rpc by default
 		AdminRPCUrl:                "http://localhost:50003",   // use a local admin rpc by default
-		TimeoutS:                   3,                          // the rpc timeout is 3 seconds
+		TimeoutS:                   30,                         // the rpc timeout is 30 seconds
```
pablocampogo: This is a global config, if the blobs rpc request needs this bigger timeout we should add it as an extra optional field and leave the general timeout as it was

fsm/indexer.go
```diff
@@ -469,6 +548,41 @@ func poolEntryKey(entry []byte) (string, error) {
 	return entryKeyOrZero(entry) // Pool.id
 }
 
+// mergeChangedBlobKeys() is changedBlobKeys' sorted-input equivalent: current and previous
+// must each be ascending by key (true for a Pebble prefix-scan result - see call sites).
+// A two-pointer merge walk finds the same added/changed/removed keys as the map-based
+// version without hashing every key into a map[string][]byte first.
+func mergeChangedBlobKeys(current, previous []blobEntry) (map[string]struct{}, map[string]struct{}) {
```
pablocampogo, line 563: This function also runs in the `IndexerBlobsFromStateChanges` path where `Current.Accounts/Previous.Accounts` come from `valuesForStateKeys(accountKeys, …)`, not an `AccountPrefix()` scan. Can we either (a) `slices.SortFunc` the account entries by key right before the merge on the journal path, or (b) keep the map-based `changedBlobKeys` for accounts on the journal path (like pools already do) and only use the merge walk where the input is a true prefix scan?
Either way a test should be added with the randomized order input.
Here the path I talked about:

- IndexerBlobsFromStateChanges builds Current/Previous via indexerBlob(..., selectiveAccounts=true), appends eventAccountKeys to the end of accountKeys, then calls DeltaIndexerBlobs(b): [fsm/indexer.go#L68-L94](https://github.com/canopy-network/canopy/blob/29dc8ee5c192cb7ba6b6a68f83917dae5b39bcdb/fsm/indexer.go#L68-L94)

- On the selectiveAccounts branch, indexerBlob fills Accounts from valuesForStateKeys(accountKeys, ...) — again appending eventAccountKeys at the end — instead of the sorted IterateAndAppend scan the legacy branch uses: [fsm/indexer.go#L158-L174](https://github.com/canopy-network/canopy/blob/29dc8ee5c192cb7ba6b6a68f83917dae5b39bcdb/fsm/indexer.go#L158-L174)

- valuesForStateKeys returns values in input-slice order and never sorts: [fsm/indexer.go#L346-L367](https://github.com/canopy-network/canopy/blob/29dc8ee5c192cb7ba6b6a68f83917dae5b39bcdb/fsm/indexer.go#L346-L367)

- DeltaIndexerBlobs passes those accounts through accountEntries straight into mergeChangedBlobKeys: [fsm/indexer.go#L418-L426](https://github.com/canopy-network/canopy/blob/29dc8ee5c192cb7ba6b6a68f83917dae5b39bcdb/fsm/indexer.go#L418-L426)

- accountEntries → entriesByKey is also order-preserving (no sort): [fsm/indexer.go#L514-L526](https://github.com/canopy-network/canopy/blob/29dc8ee5c192cb7ba6b6a68f83917dae5b39bcdb/fsm/indexer.go#L514-L526)
 
- And mergeChangedBlobKeys relies on ascending order for its c.key < p.key two-pointer comparisons: [fsm/indexer.go#L551-L584](https://github.com/canopy-network/canopy/blob/29dc8ee5c192cb7ba6b6a68f83917dae5b39bcdb/fsm/indexer.go#L551-L584)

fsm/state.go
```diff
@@ -681,6 +683,9 @@ func (s *StateMachine) IterateAndAppend(prefix []byte) (result [][]byte, err lib
 	defer it.Close()
 	// for each item of the iterator
 	for ; it.Valid(); it.Next() {
+		if cErr := ctx.Err(); cErr != nil {
+			return nil, lib.ErrCancelled(cErr)
+		}
```
pablocampogo, line 694: This check can be costly in my experience because golang inside holds a lock to check the context error. We might wan to do something like this to reduce the number of calls:
```
for i := 0; it.Valid(); it.Next() {
	if i&1023 == 0 {
		if cErr := ctx.Err(); cErr != nil {
			return nil, lib.ErrCancelled(cErr)
		}
	}
	result = append(result, it.Value())
	i++
}
```

fsm/indexer.go
```diff
@@ -71,19 +79,45 @@ func (s *StateMachine) IndexerBlobsFromStateChanges(height uint64) (b *IndexerBl
...
+	// other height -- including tip-1, the "previous" side of a live headscan request
+	// -- goes through the HSS (Historic State Store) instead. This is the actual
+	// mechanism behind steps that show a fast/slow split despite having no journal at
+	// all (validators_iterate, block_non_signers_get): it isn't proximity-based cache
+	// warmth or "how old is this height," it's this one deterministic check.
+	tier := "hss"
+	if height == s.height {
+		tier = "lss"
+	}
+	s.Metrics.RecordIndexerBlobTier(tier)
```
pablocampogo, line 122: `RecordIndexerBlobTier` fires once per `indexerBlob` call, but a legacy request builds both current and previous, so it increments twice per request — whereas path_total (incremented once per request in IndexerBlobsCached) is once. Dividing tier_total by path_total, or comparing their rates, will be off by up to 2x on the legacy path. Could we note in the canopy_indexer_blob_tier_total help text that it's counted per state-read (not per request), and/or add a path/side label so the two are reconcilable?

If these are not supposed to be retrocompatible, ignore the comment

cmd/rpc/query.go
```diff
@@ -613,25 +615,40 @@ func (s *Server) IndexerBlobs(w http.ResponseWriter, r *http.Request, _ httprout
...
+	// requestStart/path back ObserveIndexerBlobRequestTime, which times this call
+	// end-to-end for every outcome -- the number to compare against the client's
+	// observed RPC round-trip, since IndexerBlobColdReadTime and
+	// IndexerBlobJournalBuildTime each only cover one of the three branches below.
+	requestStart := time.Now()
+	path := "cache_hit"
+	defer func() {
+		s.controller.Metrics.RecordIndexerBlobPath(path)
+		s.controller.Metrics.ObserveIndexerBlobRequestTime(path, requestStart)
+	}()
```
pablocampogo, line 638: These get recorded even in the error paths. Consider using another metric for that unless this is the expected behavior

lib/metrics.go
```diff
@@ -920,6 +1105,125 @@ func (m *Metrics) UpdateStoreJobMetrics(LSScompactionTime, HSSCompactionTime, ba
...
+	m.IndexerBlobCacheSize.Set(float64(size))
+}
+
+// ObserveIndexerBlobStep() records how long a single step of an indexer-blobs read
+// took, labeled by which path (journal/legacy) the call is part of and which store
+// tier (lss/hss) served it. The tier label is what actually isolates LSS's speed
+// from HSS's independent of path/journal -- e.g. validators_iterate never has a
+// journal shortcut at all, so validators_iterate{tier="lss"} vs
+// validators_iterate{tier="hss"} is a clean tier-only comparison, unconfounded by
+// which path the call took. tier is "n/a" for steps that span two different
```
pablocampogo: It's actually `"n_a"`

scripts/deploy-node-image.sh
pablocampogo, line 1: I don't think this file should be uploaded to the repo

store/store.go
```diff
@@ -299,6 +299,10 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
 	s.MaybeCompact()
 	// backup if enabled
 	s.MaybeBackup()
+	// refresh pebble's own LSM-tree telemetry - cheap (in-memory counters only), so safe on
+	// every commit, and deliberately not gated on s.syncing like MaybeCompact() since LSM
+	// shape visibility matters most exactly when the node is catching up
+	s.metrics.UpdatePebbleMetrics(s.db.Metrics())
```
pablocampogo: This should not be ran if the node is syncing

pablocampogo (changes requested):

lib/metrics.go
```diff
@@ -72,6 +74,55 @@ type Metrics struct {
...
+	// so neither alone can account for 100% of what the client sees.
+	IndexerBlobRequestTime *prometheus.HistogramVec
+	// IndexerBlobTierTotal counts which store tier served the state read inside each
+	// indexerBlob call: "lss" (Latest State Store -- only used when the requested height
+	// equals the store's current live version) or "hss" (Historic State Store -- every
+	// other height, including tip-1). This is the real mechanism behind the "fast vs slow"
+	// split observed on steps with no journal at all (validators_iterate,
+	// block_non_signers_get) -- see store.Store.NewReadOnly's `s.version == queryVersion`
+	// check -- not a proxy like request source or distance-from-tip.
+	IndexerBlobTierTotal *prometheus.CounterVec
```
rem1niscence: These comments are too big and hard to digest, all the other metrics have a one line sentence explaining what it does on the side, follow the same pattern.

scripts/deploy-node-image.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
```
rem1niscence, line 1: Please remove this script entirely, it is useful but should not be on the canopy repository as we do not offer k8s support on this repository, it could be added to the [node](https://github.com/canopy-network/node) as that one is more akin to handle deployments specifically

store/store.go
```diff
@@ -299,6 +299,10 @@ func (s *Store) Commit() (root []byte, err lib.ErrorI) {
 	s.MaybeCompact()
 	// backup if enabled
 	s.MaybeBackup()
+	// refresh pebble's own LSM-tree telemetry - cheap (in-memory counters only), so safe on
```
rem1niscence, line 302: comment too long, only mention what it does and keep it at most one sentence as is done on the surrounding, no need to explain the **why** it does that way in detail only **what** it does, it clutters the understanding of the code itself and requires to modify the comment if the information related to it changes

store/store.go
```diff
@@ -781,8 +785,23 @@ func (s *Store) MaybeBackup() {
 	}()
 }
 
+// liveCompactionTimeout bounds a single MaybeCompact() prefix pass (LSS/HSS, running
+// live alongside sync/consensus) so a stuck compaction can't hold the single-flight
+// s.compaction lock indefinitely and starve subsequent periodic compactions.
+// TODO: per-prefix budget was chosen arbitrarily, update once multiple tests are run
+const liveCompactionTimeout = 3 * time.Minute
+
+// postSyncCompactionTimeout bounds CompactAll()'s one-time post-sync pass, which
```
rem1niscence, line 794: comment too long and talks about an specific chain which, a simpler comment like `// in larger chains, compaction needs to take more time to process the data` conveys the same explanation, does not rely on specific external information and if `CompactAll` ever changes or the caller is removed, there's no need to modify this comment

store/store.go
```diff
@@ -781,8 +785,23 @@ func (s *Store) MaybeBackup() {
 	}()
 }
 
+// liveCompactionTimeout bounds a single MaybeCompact() prefix pass (LSS/HSS, running
```
rem1niscence, line 788: comment too long, relies on the implementation of other parts of the code (`s.compaction`), keep it one line with maybe only the TODO, the variable name itself is descriptive enough to understand what it does

store/store_test.go
```diff
@@ -404,6 +404,28 @@ func validateIterators(t *testing.T, prefix string, expectedKeys []string, itera
 	}
 }
 
+// TestCompactionTimeoutsDistinct guards against collapsing the live (MaybeCompact, runs
+// inline with sync/consensus) and post-sync (CompactAll, one-time background pass that
+// must cover the indexer prefix - see store.go's postSyncCompactionTimeout doc comment)
+// compaction budgets back into a single shared constant.
+func TestCompactionTimeoutsDistinct(t *testing.T) {
+	require.Greater(t, postSyncCompactionTimeout, liveCompactionTimeout)
+}
```
rem1niscence, line 413: Unnecessary test, please remove

Makefile
```diff
@@ -112,6 +112,12 @@ docker/up-fast:
 docker/logs:
 	$(DOCKER_COMPOSE_CMD) -f $(DOCKER_DIR) logs -f --tail=1000
 
+## docker/deploy-node: build the node image and import it directly onto a k3s node's
+## containerd (no registry), then patch the matching StatefulSet to roll it out.
+## See scripts/deploy-node-image.sh for configuration env vars (NODE, NAMESPACE, etc).
+docker/deploy-node:
+	./scripts/deploy-node-image.sh
```
rem1niscence, line 119: Remove this along with the script, the comment on removing the script gives more detail on why

store/store_test.go
```diff
@@ -404,6 +404,28 @@ func validateIterators(t *testing.T, prefix string, expectedKeys []string, itera
...
+
+// TestCompactAllSucceedsOnAllPrefixes is a smoke test for CompactAll() wiring a timeout
+// through to Compact() for every store prefix, including indexerPrefix (never touched by
+// the live MaybeCompact() path).
```
rem1niscence, line 417:
```suggestion
```
Remove the comment, the test name is enough and the last part of the comment assumes that is never touched by the live `MaybeCompact()` path but it can change in the future, which would require to update this

lib/metrics.go
```diff
@@ -206,6 +257,20 @@ type StoreMetrics struct {
...
+	PebbleDiskUsageBytes      prometheus.Gauge     // total on-disk size of all sstables
+	PebbleCompactionCount     prometheus.Gauge     // total pebble-initiated compactions since process start
+	PebbleCompactionDebtBytes prometheus.Gauge     // estimated bytes still needing compaction to reach a stable LSM state
+	PebbleFlushCount          prometheus.Gauge     // total memtable flushes since process start
+	PebbleMemtableSizeBytes   prometheus.Gauge     // bytes held in memtables (incl. large flushable batches)
+	PebbleBlockCacheSizeBytes prometheus.Gauge     // bytes held in the block cache
+	PebbleBlockCacheHits      prometheus.Gauge     // cumulative block cache hits
+	PebbleBlockCacheMisses    prometheus.Gauge     // cumulative block cache misses
+	PebbleLevelTableCount     *prometheus.GaugeVec // sstable count, labeled by LSM level
+	PebbleLevelSizeBytes      *prometheus.GaugeVec // sstable bytes, labeled by LSM level
```
rem1niscence, line 273:
```suggestion PebbleReadAmp             prometheus.Gauge     // pebble's current read amplification
	PebbleDiskUsageBytes      prometheus.Gauge     // total on-disk size of all sstables
	PebbleCompactionDebtBytes prometheus.Gauge     // estimated bytes still needing compaction to reach a stable LSM state
	PebbleFlushCount          prometheus.Gauge     // total memtable flushes since process start
```

To be honest, most of these metrics wont add too much relevant information to what we do, we do not rely on pebbleDB caching at all so measuring that wont help us, the sstable count and bytes do not affect our chain in any way and  `PebbleMemtableSizeBytes` does not affect us as the memtable is periodically flushed on every block commit. I left on the suggestion only the metrics that could be useful to us at the moment, please delete the other ones

lib/config.go
rem1niscence: agreed, let's keep this as `3` and anybody who wants to modify it only has to modify their `config.json`

fsm/indexer.go
```diff
@@ -362,6 +435,10 @@ func DeltaIndexerBlobs(blobs *IndexerBlobs) (*IndexerBlobs, lib.ErrorI) {
 	}
 
 	// pools: changed+added in current, changed+removed in previous
+	// NOT converted to the sorted merge walk: poolEntryKey extracts Pool.Id's raw varint
+	// wire bytes, but the Pebble storage key (KeyForPool) encodes Id big-endian - the two
+	// orderings diverge at varint length boundaries (e.g. id 16383 vs 16384), so entry order
+	// here does not reliably match key order. Map-based diff stays correct regardless of order.
```
rem1niscence, line 441:
```suggestion
// Entry here does not reliably match key order. Map-based diff stays correct regardless of order.
```
Simpler comment without needing to dive into other implementation internals

fsm/indexer.go
```diff
@@ -342,6 +412,9 @@ func DeltaIndexerBlobs(blobs *IndexerBlobs) (*IndexerBlobs, lib.ErrorI) {
 	previous := nilSafeBlob(out.Previous)
 
 	// accounts: changed+added in current, changed+removed in previous
+	// currentAccounts/previousAccounts come from a Pebble prefix scan over AccountPrefix(),
+	// so both are already ascending by address bytes - accountEntryKey extracts that same
+	// address as the raw key, so entry order matches key order and the merge walk is safe.
```
rem1niscence, line 417:
```suggestion
```
Previous comment was clear enough

fsm/indexer.go
```diff
@@ -71,19 +79,45 @@ func (s *StateMachine) IndexerBlobsFromStateChanges(height uint64) (b *IndexerBl
...
 	if height == 0 || height > s.height {
 		height = s.height
 	}
+	// store.Store.NewReadOnly routes state reads through the LSS (Latest State Store)
+	// only when the requested height equals the store's current live version; every
+	// other height -- including tip-1, the "previous" side of a live headscan request
+	// -- goes through the HSS (Historic State Store) instead. This is the actual
+	// mechanism behind steps that show a fast/slow split despite having no journal at
+	// all (validators_iterate, block_non_signers_get): it isn't proximity-based cache
+	// warmth or "how old is this height," it's this one deterministic check.
```
rem1niscence, line 115:
```suggestion
// used for metrics to differentiate on which partition it is being worked on
```
This is explaining the whole HSS/LSS mechanism, which is also explained at the beginning of `store/store.go`. No Need to re explain it here in detail. The reader at this point should understand the differences of hss/lss

fsm/indexer.go
```diff
@@ -71,19 +79,45 @@ func (s *StateMachine) IndexerBlobsFromStateChanges(height uint64) (b *IndexerBl
...
+	s.Metrics.ObserveIndexerBlobStep("delta_compute", "journal", "n_a", deltaComputeStart)
 	return b, true, err
 }
 
-func (s *StateMachine) indexerBlob(height uint64, accountKeys [][]byte, selectiveAccounts, includeBlockEventAccounts bool) (b *IndexerBlob, err lib.ErrorI) {
+func (s *StateMachine) indexerBlob(ctx context.Context, height uint64, accountKeys [][]byte, selectiveAccounts, includeBlockEventAccounts bool) (b *IndexerBlob, err lib.ErrorI) {
+	// path labels every ObserveIndexerBlobStep call below so journal-sourced and
+	// legacy-sourced calls to the same step (accounts_iterate above all) are no
+	// longer indistinguishable in the aggregate -- selectiveAccounts is exactly the
+	// journal/legacy branch signal already threaded through this function.
```
rem1niscence, line 101:
```suggestion
// used for metrics 
```

fsm/indexer.go
```diff
@@ -71,19 +79,45 @@ func (s *StateMachine) IndexerBlobsFromStateChanges(height uint64) (b *IndexerBl
...
+		b.Previous, err = s.indexerBlob(ctx, height-1, accountKeys, true, false)
 		if err != nil {
 			return nil, true, err
 		}
 	}
+	// Previously uninstrumented: query.go's delta_compute timer only ever wrapped the
+	// legacy branch's DeltaIndexerBlobs call, leaving the journal path's own delta
+	// cost invisible. Labeling this "journal" is what actually answers "how good is
+	// the journal path" instead of assuming it from the legacy-only number.
+	deltaComputeStart := time.Now()
```
rem1niscence, line 91:
```suggestion
// Measure the time of indexing the delta differences
```

fsm/indexer.go
```diff
@@ -12,32 +14,38 @@ import (
...
 }
 
 // IndexerBlobsFromStateChanges builds an account-sparse delta using the state
 // keys journaled for height. available=false means the height predates the
 // journal and callers must use the legacy full-snapshot comparison.
-func (s *StateMachine) IndexerBlobsFromStateChanges(height uint64) (b *IndexerBlobs, available bool, err lib.ErrorI) {
+func (s *StateMachine) IndexerBlobsFromStateChanges(ctx context.Context, height uint64) (b *IndexerBlobs, available bool, err lib.ErrorI) {
+	// Honor an already-cancelled/disconnected caller before doing any store work --
+	// this path is fast, not free, and skips the per-iteration ctx checks that
+	// IterateAndAppend relies on in the legacy full-snapshot path below.
```
rem1niscence, line 45:
```suggestion
// Honor an already-cancelled/disconnected caller before doing any store work 
```
The gold rule of thumb, comments at the function level should, most of the time, explain **what** not **why**, the why is left for top level entities like at the top of a function, logical module or package

rem1niscence (changes requested):
Most of the comments, if not all, are related to reducing noise in the code, overly explanations related to ultra specific use cases will be forgotten over time and can confuse readers, specially new ones in the future, I'd highly suggest do a thorough check of the comments here to minimize as much as possible and only keep the relevant information available

PR #506, opened 2026-08-07
feat(indexer): sparse validators and non-signers in the state-change journal

ezeike:
## Summary

Extends the indexer-blob state-change journal (previously account-only sparse deltas) to also cover validators and non-signers, replacing PR 502 (`perf/indexer-journal-entities`), which had correctness bugs this branch fixes:

- **Journal restructure**: one row per touched key → one row per `(version, entity-type)` bucket, cutting per-key storage overhead and turning prefix-filtered reads into a single point read.
- **Non-signer address bug fixed**: addresses are now recovered from the storage key instead of relying on the (never-carried) value — the old journal path silently returned non-signer deltas with empty addresses.
- **Validator force-inclusion fixed**: reward/slash/lifecycle-event validators are now force-included directly by key (the event already names the key) — no index needed, unlike PR 502's approach.
- **Validator/delegate status totals** moved from a per-commit full scan (~200ms at current validator+delegate counts, ~44,000 combined) into blob-build time, computed incrementally from a cached baseline with a full-scan fallback.

## Notable mid-implementation findings

- The original design's disk-persisted totals baseline never actually persisted — `GetValidatorTotals`/`SetValidatorTotals` were called through a `TimeMachine()` read-only snapshot whose writes are silently discarded. Replaced with an in-memory, LRU-bounded, per-`Store`-shared cache instead (totals are fully re-derivable via the existing full-scan fallback, so this trades disk-durability for simplicity with no correctness gap).
- Added `singleflight`-based dedup so concurrent RPC requests racing on the same missing cache entry don't each pay a redundant full scan.

## Deploy ordering note

This sets `non_signers_delta: true` on sparse-path blobs. **Do not enable `StateChangeJournalEnabled` on a fleet until `canopy-indexer`'s companion PR (honoring `non_signers_delta`) is deployed** — otherwise the indexer will misfire spurious "reset" rows for untouched non-signers. `StateChangeJournalEnabled` defaults `false` today, so this is not an active bug, just a rollout-sequencing requirement.


fsm/indexer.go
```diff
@@ -185,7 +210,16 @@ func (s *StateMachine) indexerBlob(ctx context.Context, height uint64, accountKe
 	}
 	// retrieve validators
 	stepStart = time.Now()
-	validators, err := sm.IterateAndAppend(ctx, ValidatorPrefix())
+	var validators [][]byte
+	if !p.selective {
+		validators, err = sm.IterateAndAppend(ctx, ValidatorPrefix())
+	} else {
+		forced, forceErr := validatorForceKeysByAddress(blockBz)
```
pablocampogo: validatorForceKeysByAddress(blockBz) gets called twice per Current blob — here, and again at [indexer.go#L384](https://github.com/canopy-network/canopy/blob/884da7039dfa5ab43957f09f110840bd416393fd/fsm/indexer.go#L384) to build touchedKeys. Both paths run for the Current blob (selective + includeBlockEventAccounts are both true), and each call does a full lib.Unmarshal of the entire block inside the [helper](https://github.com/canopy-network/canopy/blob/884da7039dfa5ab43957f09f110840bd416393fd/fsm/indexer.go#L801-L809), not just an event walk.
Compute `forced` once after `blockBz` is available and reuse it at both L221 and L388 to reduce substantially the number of unmarshals

pablocampogo (changes requested):

PR #507, opened 2026-08-07
cache LoadCommittee's result for blockNonSignerAddresses

ezeike:
Stacked on #506 (targets `indexer-blob-journal-validators-nonsigners`, not `main` — merge order matters).

## Problem

`block_non_signers_get` was the dominant remaining cost on the indexer-blob journal path (~541ms hss + ~160ms lss per request, confirmed on staging) — none of #506's journal work touched it, since `blockNonSignerAddresses` derives committee membership from QC/committee data via `LoadCommittee`, not the state-change journal. `LoadCommittee` unconditionally re-scans the full validator/delegate set (~44k) on every call.

## Fix

Memoization cache (`GetOrComputeCommittee`, mirrors `GetOrComputeValidatorTotals`'s LRU+singleflight shape) keyed by `(chainId, rootHeight)`, populated lazily from the real, unmodified `LoadCommittee` on a miss.

**No incremental diffing** — `ValidatorSet.MultiKey` is a BLS-aggregated point over every member; incrementally maintaining that is real cryptographic surface deliberately avoided here. Every miss pays `LoadCommittee`'s full cost, same as today.

**Scoped to the indexer read path only** — a new `cachedLoadCommittee` wrapper, called only from `blockNonSignerAddresses`. `LoadCommittee` itself is unmodified; its six other, consensus-critical callers (bft message/evidence validation, certificate-results handling, validator-set transitions) are unaffected and never see this cache.

## Steady-state effect

A request's `Previous` call resolves at the same `(chainId, rootHeight)` the prior request's `Current` call already computed and cached one block earlier, collapsing two full computations into one — independent of which tier (hss/lss) either call happens to land on.

Same hit/miss/dedup/size metric set as the validator-totals cache.


fsm/state.go
```diff
@@ -467,6 +473,26 @@ func (s *StateMachine) LoadCommittee(chainId uint64, height uint64) (lib.Validat
...
+		committee, loadErr := s.LoadCommittee(chainId, rootHeight)
+		if loadErr != nil {
+			return nil, loadErr
+		}
+		return &committee, nil
+	})
+	if err != nil {
+		return lib.ValidatorSet{}, err
+	}
+	return *vs, nil
```
pablocampogo: You should return a copy here not directly the pointer. Changes in this pointer can cause changes in the unintentional changes in the cache

store/indexer.go
```diff
@@ -203,6 +232,40 @@ func (t *Indexer) GetOrComputeValidatorTotals(version uint64, compute func() (*l
...
 }
 
+// GetOrComputeCommittee returns (chainId, rootHeight)'s cached committee, computing via
+// compute exactly once even under concurrent callers, then caching the result.
+func (t *Indexer) GetOrComputeCommittee(chainId, rootHeight uint64, compute func() (*lib.ValidatorSet, lib.ErrorI)) (*lib.ValidatorSet, lib.ErrorI) {
+	if vs, ok := t.committees.get(chainId, rootHeight); ok {
+		t.metrics.RecordCommitteeCacheHit()
+		return vs, nil
+	}
+	t.metrics.RecordCommitteeCacheMiss()
```
pablocampogo: This should be counted inside the `do` function after the recheck

store/committee_singleflight_test.go
pablocampogo, line 1: Just as good measure add a test that checks that asserts the error path is not cached

pablocampogo (changes requested):

PR #512, opened 2026-08-11
fix: make cli params go to binary with auto update

pablocampogo:
## Summary
Make `--data-dir` (and the other global flags `--rpc-url` / `--admin-url`) work end-to-end with the auto-updater and Docker. Previously they were honored only by the standalone CLI; under the auto-updater everything fell back to `~/.canopy`.

## Changes
- **cli**: extract shared flag registration; add `ParseGlobalFlags` (parse flags outside cobra) and `GlobalFlagArgs` (reconstruct flags for the child); pass the resolved data dir to the logger so logs land in the custom dir.
- **auto-update**: parse global flags on startup, resolve the updater's own config against them, and forward them to the child process (`start`) and the `version` check.
- **entrypoint.sh**: resolve the data dir from `--data-dir` (or `DATA_DIR`) instead of hardcoding `/root/.canopy`.
- **tests**: unit tests for `ParseGlobalFlags` / `GlobalFlagArgs`.

### DIFF
--- a/cmd/auto-update/coordinator.go
+++ b/cmd/auto-update/coordinator.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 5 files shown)


cmd/auto-update/coordinator.go
```diff
@@ -26,16 +27,18 @@ type Supervisor struct {
 	exit           chan error           // channel to notify listeners when process exits
 	unexpectedExit chan error           // channel to notify listeners when process exits unexpectedly
 	pluginConfig   *PluginReleaseConfig // optional plugin configuration
+	childArgs      []string             // global flags forwarded to the CLI child process
 	log            lib.LoggerI          // logger instance
 }
 
 // NewSupervisor creates a new ProcessSupervisor instance
-func NewSupervisor(logger lib.LoggerI, pluginConfig *PluginReleaseConfig) *Supervisor {
+func NewSupervisor(logger lib.LoggerI, pluginConfig *PluginReleaseConfig, childArgs []string) *Supervisor {
```
rem1niscence, line 35: if any childArgs can be passed arbitrarily, why on the `/auto-update/main.go` dataDirPath is passed specially to the coordinator?

pablocampogo: Because the coordinator needs the data dir for its own operations

andrewnguyen22 (approved):

PR #518, opened 2026-08-13
Add a small offline native transaction signer

andrewnguyen22:
Pretty small on purpose. This adds a native Canopy send signer that can be imported as a Go package or used as a standalone binary.

It uses the existing encrypted keystore, requires an explicit timestamp so retries stay deterministic, rejects the reserved RLP memos, and can inspect signed transactions before they are broadcast. The release workflow also publishes signer binaries and checksums.

Checked with:
- focused unit tests
- race detector
- go vet
- FSM tests
- Linux static build


pablocampogo (commented):
Maybe you want to add comments per line as the rest of the codebase?

pablocampogo (approved):

PR #523, opened 2026-08-14
feat: make release actions validate semantics before hand and add needed prefixes for plugins

pablocampogo:
## Summary

Release workflows previously accepted a free-form `tag` input and passed it straight to the GitHub release step. This allowed malformed tags (e.g. `plugin-go-v1.0.4.5`) to be published, which then broke the auto-updater since it validates versions as strict semver (`MAJOR.MINOR.PATCH`) and rejects anything else with `invalid release version`.

This PR adds up-front version validation to all release workflows and simplifies the plugin release inputs so the language prefix can no longer be mistyped.

### All release workflows (`release.yml` + 5 plugin workflows)
- Added a `validate` job that runs **before** any build/release step. If the version isn't a valid semantic version (`MAJOR.MINOR.PATCH`, with optional `-prerelease`/`+build`), the job fails fast with a clear `::error::` annotation and nothing is built or published.
- The `release` job (and, in `release.yml`, the downstream Docker jobs) now depends on `validate` via `needs`, so the whole pipeline is gated.

### Plugin workflows (`go`, `python`, `typescript`, `kotlin`, `csharp`)
- Changed the `workflow_dispatch` input from `tag` to `version`. Operators now enter only the semantic version (e.g. `1.0.0`).
- The `validate` job computes the full tag by automatically prepending the correct prefix (`plugin-<language>-v`) and exposes it as a job output.
- The `release` job consumes `needs.validate.outputs.tag` for `tag_name`, the release name, and the release body.

## Why

- **Prevents bad releases**: invalid versions fail before a GitHub release or Docker image is created, instead of surfacing later as an auto-updater error.
- **Removes a footgun**: operators can no longer forget or mistype the `plugin-<language>-` prefix — it's added automatically and consistently.
- **Consistent with the auto-updater**: the validation is intentionally stricter than `golang.org/x/mod/semver` (requires full `MAJOR.MINOR.PATCH`), so anything that passes will be accepted by the updater.

## Behavior change

- Plugin release runs now take a **version** (`1.0.4`) rather than a full **tag** (`plugin-go-v1.0.4`). The resulting tag is unchanged in shape (e.g. `plugin-go-v1.0.4`).
- The main `release.yml` still takes a `v`-prefixed tag input; only validation was added there.

### DIFF
--- a/.github/workflows/release-plugin-csharp.yml
+++ b/.github/workflows/release-plugin-csharp.yml
@@ -18,25 +18,27 @@ jobs:
       tag: ${{ steps.compute.outputs.tag }}
 
     steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.26"
+
       - name: Validate version and compute tag
         id: compute
         env:
           VERSION: ${{ github.event.inputs.tag }}
         run: |
           PREFIX="plugin-csharp-v"
-          # Backwards compatibility: also accept the fully-prefixed tag form
-          # (e.g. plugin-csharp-v1.0.4) in addition to the bare version (1.0.4).
-          # Old forks and their release automation still pass the full tag, so
-          # strip the prefix here when it's present. This is intentionally NOT
-          # listed in the workflow_dispatch input description above so the manual
-          # flow steers users toward entering just the version.
-          # DO NOT REMOVE: deleting this breaks releases triggered from existing forks.
+          # Back-compat: accept the full tag form (e.g. plugin-csharp-v1.0.4) too, not just
+          # the bare version. DO NOT REMOVE: old forks pass the full tag when releasing.
           VERSION="${VERSION#"$PREFIX"}"
-          SEMVER_REGEX='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*)(\.(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*))*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$'
-          if [[ ! "$VERSION" =~ $SEMVER_REGEX ]]; then
-            echo "::error::Version '$VERSION' is not a valid semantic version. Expected MAJOR.MINOR.PATCH (e.g. 1.0.0)"
-            exit 1
-          fi
+          # Tolerate a bare "v" prefix (e.g. v1.0.4) so the tag isn't doubled below.
+          VERSION="${VERSION#v}"
+          # Validate via the shared script (same semver check as the auto-updater).
+          go run ./cmd/scripts/semver "$VERSION"
           echo "tag=${PREFIX}${VERSION}" >> "$GITHUB_OUTPUT"
           echo "Version '$VERSION' is valid. Release tag: ${PREFIX}${VERSION}"
 
--- a/.github/workflows/release-plugin-csharp.yml
+++ b/.github/workflows/release-plugin-csharp.yml
@@ -20,8 +20,9 @@ jobs:
     steps:
       - name: Validate version and compute tag
         id: compute
+        env:
+          VERSION: ${{ github.event.inputs.version }}
         run: |
-          VERSION="${{ github.event.inputs.version }}"
           SEMVER_REGEX='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*)(\.(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*))*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$'
           if [[ ! "$VERSION" =~ $SEMVER_REGEX ]]; then
             echo "::error::Version '$VERSION' is not a valid semantic version. Expected MAJOR.MINOR.PATCH (e.g. 1.0.0)"
--- a/.github/workflows/release-plugin-csharp.yml
+++ b/.github/workflows/release-plugin-csharp.yml
@@ -3,17 +3,37 @@ name: Release C# Plugin
 on:
   workflow_dispatch:
     inputs:
-      tag:
-        description: "Tag for the plugin release (e.g. plugin-csharp-v1.0.0)"
+      version:
+        description: "Semantic version for the plugin release (e.g. 1.0.0). The 'plugin-csharp-v' prefix is added automatically."
         required: true
 
 permissions:
   contents: write
 
 jobs:
+  validate:
+    name: Validate version
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.compute.outputs.tag }}
+
+    steps:
+      - name: Validate version and compute tag
+        id: compute
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
+          SEMVER_REGEX='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$'
+          if [[ ! "$VERSION" =~ $SEMVER_REGEX ]]; then
+            echo "::error::Version '$VERSION' is not a valid semantic version. Expected MAJOR.MINOR.PATCH (e.g. 1.0.0)"
+            exit 1
+          fi
+          echo "tag=plugin-csharp-v${VERSION}" >> "$GITHUB_OUTPUT"
+          echo "Version '$VERSION' is valid. Release tag: plugin-csharp-v${VERSION}"
+
   release:
     name: Build and Release C# Plugin
     runs-on: ubuntu-latest
+    needs: validate
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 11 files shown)


.github/workflows/release.yml
```diff
@@ -3,17 +3,37 @@ name: Release
...
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.compute.outputs.tag }}
+
+    steps:
+      - name: Validate version and compute tag
+        id: compute
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
+          SEMVER_REGEX='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$'
```
andrewnguyen22: This regex is a little looser than SemVer in the prerelease/build sections. It accepts values like `1.2.3-01`, `1.2.3-..`, and `1.2.3-alpha.`, which `golang.org/x/mod/semver` rejects later. That leaves a path to the same broken release/update behavior this PR is trying to prevent. Could we use a fully conforming SemVer check here?

.github/workflows/release.yml
```diff
@@ -3,17 +3,37 @@ name: Release
...
+    name: Validate version
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.compute.outputs.tag }}
+
+    steps:
+      - name: Validate version and compute tag
+        id: compute
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
```
andrewnguyen22: Could we pass this through `env` instead of interpolating it directly into the `run` script? GitHub expands the expression before Bash parses it, so an input such as `1.2.$(printf 3)` runs the command substitution and then validates as `1.2.3`. Something like `env: VERSION: ${{ inputs.version }}` keeps the input as data rather than shell source.

.github/workflows/release.yml
```diff
@@ -3,17 +3,37 @@ name: Release
 on:
   workflow_dispatch:
     inputs:
-      tag:
-        description: "Tag for the release (e.g. v1.0.0)"
+      version:
```
andrewnguyen22: It looks like the main release workflow changed from `tag` to `version` too, but the PR description says the main workflow will keep accepting a `v`-prefixed tag. This also breaks existing calls like `gh workflow run release.yml -f tag=v1.2.3`. Was the main workflow meant to keep the old input contract? If the rename is intentional, the behavior change should be called out and any callers migrated.

andrewnguyen22 (changes requested):
Nice improvement overall — having operators enter just the version should make releases a lot harder to mistype. I found a few edge cases worth tightening up before this lands. The validation block is repeated across all six workflows, so the same fixes should apply everywhere.

.github/workflows/release.yml
pablocampogo: I just did it for consitency, and this action at the moment is always ran manually

.github/workflows/release.yml
pablocampogo: Fixed

.github/workflows/release.yml
pablocampogo: Fixed

.github/workflows/release-plugin-csharp.yml
```diff
@@ -4,16 +4,46 @@ on:
...
+          VERSION: ${{ github.event.inputs.tag }}
+        run: |
+          PREFIX="plugin-csharp-v"
+          # Backwards compatibility: also accept the fully-prefixed tag form
+          # (e.g. plugin-csharp-v1.0.4) in addition to the bare version (1.0.4).
+          # Old forks and their release automation still pass the full tag, so
+          # strip the prefix here when it's present. This is intentionally NOT
+          # listed in the workflow_dispatch input description above so the manual
+          # flow steers users toward entering just the version.
+          # DO NOT REMOVE: deleting this breaks releases triggered from existing forks.
```
rem1niscence: is it really necessary to support such existing releases? is better to go into mainnet with a fresh setup and require the existing chains to update their versioning as mainnet is a major migration. I'd rather get rid of this backwards compatibility. Same comment for all the other files that have this

.github/workflows/release-plugin-csharp.yml
```diff
@@ -4,16 +4,46 @@ on:
...
+          VERSION="${VERSION#"$PREFIX"}"
+          SEMVER_REGEX='^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*)(\.(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*))*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$'
```
rem1niscence: I'd assume this regex was generated using AI but ultimately is complex and right now we can't guarantee it will support every valid semver combination, this is using `bash` but why not better remake it using `actions/setup-go@v7` that way you can make a tiny go script that calls `semver.IsValid({input})` so we can always guaranteed the semver will work without resorting into regex but the actual source of truth

rem1niscence (changes requested):

.github/workflows/release-plugin-csharp.yml
andrewnguyen22: Your call - I'll retract my ask of backwards compatibility

pablocampogo: I'll go back to delete this after the change in the graduator side, since we're making them update their forks. That was my original plan anyways

.github/workflows/release-plugin-csharp.yml
pablocampogo: Changed so both the auto update and the action use the same code to validate

lib/util.go
```diff
@@ -1093,3 +1094,17 @@ func RandSlice(byteSize uint64) []byte {
...
+	return semver.IsValid(normalizeVersion(version))
+}
+
+// normalizeVersion adds the leading "v" that golang.org/x/mod/semver requires.
+func normalizeVersion(version string) string {
+	if version == "" || version[0] == 'v' {
+		return version
+	}
+	return "v" + version
+}
```
rem1niscence, line 1110: I Think this could be removed, the only thing this does under the hood is to use golang's semver and it wont change, this util is not needed as we can call the semver function directly

cmd/scripts/semver/main_test.go
```diff
@@ -0,0 +1,29 @@
...
+	t.Run("empty version", func(t *testing.T) {
+		require.Error(t, run([]string{""}))
+	})
+	t.Run("no arguments", func(t *testing.T) {
+		require.Error(t, run(nil))
+	})
+	t.Run("too many arguments", func(t *testing.T) {
+		require.Error(t, run([]string{"1.0.0", "2.0.0"}))
+	})
+}
```
rem1niscence, line 29: This whole test suite can be removed, again, is calling a golang's native package, native packages should not be tested as they're already tested on its own code, it makes this reduntandt

cmd/scripts/semver/main.go
```diff
@@ -0,0 +1,30 @@
...
+// run validates the single version argument.
+func run(args []string) error {
+	if len(args) != 1 || args[0] == "" {
+		return fmt.Errorf("usage: semver <version>")
+	}
+	if !lib.IsValidVersion(args[0]) {
+		return fmt.Errorf("invalid semantic version: %q (expected MAJOR.MINOR.PATCH, e.g. 1.2.3 or v1.2.3)", args[0])
+	}
+	return nil
+}
```
rem1niscence, line 30: the helper added the `v` automatically but I think we should not, the v should be added by the user so they're aware of the semantic versioning used, otherwise will fail, they retry again and that's it

cmd/auto-update/releaser.go
```diff
@@ -211,16 +212,16 @@ func (rm *ReleaseManager) ShouldUpdate(release *Release) error {
 	candidate := semver.Canonical(candidateTag)
 	current := semver.Canonical(rm.Version)
 	// for plugins, if current version is invalid (first run), always update
-	if rm.config.Type == ReleaseTypePlugin && (current == "" || !semver.IsValid(current)) {
+	if rm.config.Type == ReleaseTypePlugin && (current == "" || !lib.IsValidVersion(current)) {
```
rem1niscence, line 215: revert these to `semver.IsValid`, again, no other things are done beside calling this under the hood, better to call the function directly

rem1niscence (changes requested):

cmd/auto-update/releaser.go
pablocampogo: This was done exactly so both actions and auto update use the same function. I'll keep it

cmd/scripts/semver/main.go
pablocampogo: That v has caused problem with some users, I left everything so the user could input or not on purpose.

lib/util.go
pablocampogo: This is used so both action and auto update use exactly the same function

cmd/scripts/semver/main_test.go
pablocampogo: This is to prevent breaking this script in the future, because the CI wouldn't run and this script has become a really important part of our infrastructure

cmd/scripts/semver/main.go
rem1niscence: disagreed, the guide must say that users should add that and otherwise will fail, there should be requirements that must be followed

cmd/scripts/semver/main_test.go
rem1niscence: still unnecessary, my opinion still stands, go changes are guaranteed to be non breaking, and when breaking changes are done they are changed to a v2 packages, check https://go.dev/blog/v2-go-modules

lib/util.go
rem1niscence: and both only call a single line which is `semvr.IsValid`, what's the difference?

cmd/scripts/semver/main.go
pablocampogo: It doesn't hurt to accept both, doesn't cost us anything and in a future reduces support ticket calls

lib/util.go
pablocampogo: In a future we might add something there, the unit tests would catch a diverge when they run and we stopped an issue.
That's the main advantage of this go script

rem1niscence (approved):

rem1niscence:
:)

PR #525, opened 2026-08-18
Feat/auto update binary persistence

rem1niscence:
## Description

Currently, when the canopy auto updater downloaded a new binary in a containerized environment such as docker, it was saved in a non-volume path that would be overwritten with the original file on restarts, this meant that everytime the canopy binary was restarted, it would start at an older version, more specifically, the version of the auto updater.

This fix downloads updates to the `data-dir` folder and prefers to execute the canopy binary from there, if not found, it will use the original one from the auto updater.

## Related Issues

Closes: N/A

## Changes Made

- Fix bug where restarting canopy containers would revert the new binary version
- Remove unnecessary `releaser.Enabled` field

## Checklist
- [X] I have tested the changes locally and verified they work as intended.
- [ ] I have appropriately titled my branch `issue-#<issue-number>`.
- [ ] I have run re-built the web-wallet and/or block explorer (if applicable).
- [ ] I have run `npm run prettier` to format the web-wallet and/or block explorer (if applicable)

### DIFF
--- a/cmd/auto-update/main.go
+++ b/cmd/auto-update/main.go
@@ -26,6 +26,7 @@ const (
 	defaultRepoName     = "canopy"
 	defaultRepoOwner    = "canopy-network"
 	defaultBinPath      = "./cli"
+	defaultNewBinName   = "canopy_updated" // name of the downloaded binary, kept on the data directory
 	defaultCheckPeriod  = time.Minute * 30 // default check period for updates
 	defaultGracePeriod  = time.Second * 2  // default grace period for graceful shutdown
 	defaultMaxDelayTime = 30               // default max delay time for staggered updates (minutes)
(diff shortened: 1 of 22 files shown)


pablocampogo (changes requested):
Let me test these changes with my latest change in the graduator and add plugins to the mix

pablocampogo (approved):

controller/controller.go
```diff
@@ -278,6 +278,10 @@ func (c *Controller) runPluginCtl(plugin, action string) ([]byte, lib.ErrorI) {
 	}
 	// create the command using the requested action
 	cmd := exec.Command(cmdPath, action)
+	// point the launcher at the persistent plugin home under the data dir so
+	// artifacts survive restarts (must match the coordinator's PluginDir)
```
rem1niscence:
```suggestion
	// point the launcher at the persistent plugin home under the data dir so
	// artifacts survive restarts
```
The controller should not be aware of the coordinator but the other way around, is the coordinators responsibility to match the controller's behavior

cmd/auto-update/main.go
```diff
@@ -227,7 +227,7 @@ func getConfigs() (*Configs, lib.LoggerI) {
 				Type:           ReleaseTypePlugin,
 				RepoOwner:      repoOwner,
 				RepoName:       repoName,
-				PluginDir:      fmt.Sprintf("plugin/%s", canopyConfig.Plugin),
+				PluginDir:      filepath.Join(canopyConfig.DataDirPath, "plugin", canopyConfig.Plugin),
```
rem1niscence: This mentions the dir as the `/{DataDirPath/plugin...` but on the the dockerfiles of kotlin and C# I see them being built in `/app/plugin` why dont you use the same approach as the canopy binary to have a `build` folder where the original binary is located and an `updates` folder where updates go in the `DataDir` as otherwise I think this code will fail for some plugins as in the will either not be found or suffer the same bug as starting with a previous version from the Dockerfile

pablocampogo: Fixed the issue for the Dockerfiles of the plugin without auto update, this unmask a bigger bug regarding versions, so good catch

pablocampogo: That being set, using a different dir for updates in the plugin case for me it's not worth it. In all the languages but go these are actually not binaries and they would require  a lot of setup, the graduator can handle that as other deployment stacks

controller/controller.go
pablocampogo: Done

PR #535, opened 2026-08-21
fix(controller): guard against nil account in UpdateTelemetry

erhnysr:
## Problem

`UpdateTelemetry()` in `controller/block.go:775` drops the error from `GetAccount()` and dereferences the returned account directly:

```go
// update account metrics
if a, _ := c.FSM.GetAccount(address); a.Amount != 0 {
	c.Metrics.UpdateAccount(address.String(), a.Amount)
}
```

`GetAccount()` returns a **nil** account together with an error on a store read or unmarshal failure (`fsm/account.go:31-35`):

```go
bz, err := s.Get(KeyForAccount(address))
if err != nil {
	return nil, err
}
acc, err := s.unmarshalAccount(bz)
if err != nil {
	return nil, err
}
```

So when the store errors, `a` is `nil` and `a.Amount` panics with a nil pointer dereference. `UpdateTelemetry()` runs in a `defer` after the block is committed, so the panic surfaces on the post-commit path.

The core argument is an **internal inconsistency**: the validator branch immediately above already guards its result, while the account branch does not:

```go
// update validator metric
if v, _ := c.FSM.GetValidator(address); v != nil && v.StakedAmount != 0 {   // guarded
	...
}
// update account metrics
if a, _ := c.FSM.GetAccount(address); a.Amount != 0 {                       // not guarded
	...
}
```

`GetValidator()` and `GetAccount()` share the same `(nil, err)`-on-failure contract, so the two branches should be guarded the same way.

## Fix

A single clause, matching the validator branch directly above:

```go
if a, _ := c.FSM.GetAccount(address); a != nil && a.Amount != 0 {
```

Returning the error is not an option here: `UpdateTelemetry()` is a `void` best-effort telemetry function, and the in-function convention is drop-the-error, guard-the-result (exactly what the validator branch does). Propagating the error would mean changing the signature and every caller for a metrics-only side effect, which is out of scope and inconsistent with the surrounding code.

## Tests

New test in `controller/block_test.go` that injects a store whose `Get()` always fails, forcing `GetAccount()` to return `(nil, err)`, then asserts `UpdateTelemetry()` does not panic. This follows the failing-store injection pattern from merged PR #530, with two deliberate differences:

- It asserts `require.NotPanics` rather than #530's `require.ErrorContains`, because this symptom is a panic, not a returned error (`UpdateTelemetry()` returns nothing).
- I could not reuse #530's in-package `newTestStateMachine`, which constructs a `StateMachine` by setting unexported fields directly (only legal inside the `fsm` package). From the `controller` package I used the exported `SetStore()` plus minimal reflection to allocate the unexported `cache` field — the same cross-package construction pattern the `cmd/rpc` tests already use.

### Failing first — before the fix (`cdfd182`)

```
=== RUN   TestUpdateTelemetryHandlesGetAccountError
    block_test.go:44:
        	Error Trace:	/Users/macbookpro/canopy/controller/block_test.go:44
        	Error:      	func (assert.PanicTestFunc)(0x101493f40) should not panic
        	            		Panic value:	runtime error: invalid memory address or nil pointer dereference
        	            		Panic stack:	goroutine 38 [running]:
        	            	...
        	            	panic({0x10214f9a0?, 0x102377d60?})
        	            		/opt/homebrew/Cellar/go/1.26.3/libexec/src/runtime/panic.go:860 +0x12c
        	            	github.com/canopy-network/canopy/controller.(*Controller).UpdateTelemetry(...)
        	            		/Users/macbookpro/canopy/controller/block.go:775 +0x1ec
        	            	github.com/canopy-network/canopy/controller.TestUpdateTelemetryHandlesGetAccountError.func1()
        	            		/Users/macbookpro/canopy/controller/block_test.go:45 +0x28
        	            	...
        	Test:       	TestUpdateTelemetryHandlesGetAccountError
--- FAIL: TestUpdateTelemetryHandlesGetAccountError (0.00s)
FAIL
FAIL	github.com/canopy-network/canopy/controller	0.650s
FAIL
```

The panic stack points straight at `controller/block.go:775`.

### Passing — after the fix (`6b95794`)

```
=== RUN   TestUpdateTelemetryHandlesGetAccountError
--- PASS: TestUpdateTelemetryHandlesGetAccountError (0.00s)
PASS
ok  	github.com/canopy-network/canopy/controller	0.540s
```

Full package run:

```
ok  	github.com/canopy-network/canopy/controller	0.438s
```

## Notes

`CONTRIBUTING.md` mentions updating `.docs/CHANGELOG.md`, but that file does not exist in the repository and recent merged PRs (e.g. #494) did not add one, so I did not fabricate a changelog entry.

### DIFF
--- a/controller/block.go
+++ b/controller/block.go
@@ -772,7 +772,7 @@ func (c *Controller) UpdateTelemetry(qc *lib.QuorumCertificate, block *lib.Block
 		c.Metrics.UpdateValidator(address.String(), v.StakedAmount, v.UnstakingHeight != 0, v.MaxPausedHeight != 0, v.Delegate, v.Compound, isProducer, nonSigners, doubleSigners)
 	}
 	// update account metrics
-	if a, _ := c.FSM.GetAccount(address); a.Amount != 0 {
+	if a, _ := c.FSM.GetAccount(address); a != nil && a.Amount != 0 {
 		c.Metrics.UpdateAccount(address.String(), a.Amount)
 	}
 }
--- a/.github/workflows/protect-main.yaml
+++ b/.github/workflows/protect-main.yaml
@@ -1,0 +1,26 @@
+name: Protect main
+
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - edited
+
+jobs:
+  require-development:
+    name: Require development source branch
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check source branch
+        env:
+          HEAD_REF: ${{ github.head_ref }}
+        run: |
+          if [ "$HEAD_REF" != "development" ]; then
+            echo "::error::PRs targeting main must come from the development branch (got '$HEAD_REF'). Open this PR against development instead, then promote development to main."
+            exit 1
+          fi
+          echo "PR is from development → main. OK."
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,9 +84,10 @@ discuss your intended approach for solving the problem in the comments for an ex
 - **Update the CHANGELOG** for all enhancements and bug fixes. Include the corresponding date, issue
   number if one exists and current version if any. Check the format of the [CHANGELOG](.docs/CHANGELOG.md).
 
-- **Use the repo's default main branch.** Branch from and
+- **Target the `development` branch.** The repo's default branch is still `main`, but do **not**
+  open pull requests against `main`. Branch from and
   [submit your pull request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork)
-  to the repo's default branch `main`.
+  to `development`.
 
 -
   **[Resolve any merge conflicts](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-on-github)**
--- a/controller/block_test.go
+++ b/controller/block_test.go
@@ -1,0 +1,68 @@
+package controller
+
+import (
+	"bytes"
+	"reflect"
+	"sync/atomic"
+	"testing"
+	"unsafe"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/canopy-network/canopy/fsm"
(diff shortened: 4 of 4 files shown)


erhnysr:
@pablocampogo the latest CI run on this branch is stuck at `action_required` rather than pass/fail — looks like it just needs a maintainer to approve running the workflow for this PR (outside-contributor gate), not a code issue. Could you approve it from the Actions tab when you get a chance? Fix itself (nil-account guard in UpdateTelemetry) hasn't changed since it was opened.

controller/block.go
```diff
@@ -780,7 +780,7 @@ func (c *Controller) UpdateTelemetry(qc *lib.QuorumCertificate, block *lib.Block
 		c.Metrics.UpdateValidator(address.String(), v.StakedAmount, v.UnstakingHeight != 0, v.MaxPausedHeight != 0, v.Delegate, v.Compound, isProducer, nonSigners, doubleSigners)
 	}
 	// update account metrics
-	if a, _ := c.FSM.GetAccount(address); a.Amount != 0 {
+	if a, _ := c.FSM.GetAccount(address); a != nil && a.Amount != 0 {
```
pablocampogo: Great catch! But please add an error log for when account is empty to there's visibility of the error.

pablocampogo (changes requested):

erhnysr:
@pablocampogo added in d21a689 - when GetAccount returns nil, UpdateTelemetry now logs it via c.log.Errorf with the address, instead of silently skipping the metric update. Amount == 0 (a real, funded-then-emptied account) still skips silently as before, only the nil case is now logged.

pablocampogo:
Tests failed @erhnysr

erhnysr:
@pablocampogo fixed in 6604eed - the new error log in UpdateTelemetry panicked in TestUpdateTelemetryHandlesGetAccountError because that test's minimal Controller never set the log field (nil LoggerI, calling .Errorf on it panics). Wired lib.NewNullLogger() into the test the way a real Controller always has a logger set. Full controller package is green now (10/10 tests passing), including the target test.

pablocampogo (approved):

PR #538, opened 2026-08-23
fix: bound inbound transaction spam resources

andrewnguyen22:
A few small changes to make inbound transaction spam less painful:

- turned the inbound P2P bandwidth limiter back on
- changed the message dedupe cache to keep hashes instead of full messages
- capped the failed transaction cache at 5,000 normal-sized transactions
- oversized failures still get removed from the mempool, they just won’t hang around for RPC display

This doesn’t change transaction validity, fee ordering, or consensus behavior.

Tested with `go test ./lib ./controller ./p2p`.

### DIFF
--- a/lib/mempool.go
+++ b/lib/mempool.go
@@ -13,6 +13,11 @@ import (
 
 var _ Mempool = &FeeMempool{} // Mempool interface enforcement for FeeMempool implementation
 
+const (
+	failedTxCacheMaxEntries = 5000
+	failedTxCacheMaxTxBytes = 4 * 1024
+)
+
 // Mempool interface is a model for a pre-block, in-memory, Transaction store
 type Mempool interface {
 	Contains(txHash string) bool                             // whether the mempool has this transaction already (de-duplicated by hash)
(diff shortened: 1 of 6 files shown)


lib/mempool.go
```diff
@@ -376,6 +381,15 @@ func (f *FailedTxCache) Add(failed *FailedTx) (added bool) {
 	f.l.Lock()
 	// unlock when the function completes
 	defer f.l.Unlock()
+	if failed == nil || len(failed.bytes) > failedTxCacheMaxTxBytes {
+		return false
+	}
+	if _, exists := f.cache[failed.Hash]; !exists && len(f.cache) >= failedTxCacheMaxEntries {
+		for hash := range f.cache {
+			delete(f.cache, hash)
```
rem1niscence, line 389: why not recreate the  `f.cache` variable? is faster (leaves the cleanup to the GC) and go maps are known for keeping data in memory even after deleting all the items

rem1niscence (approved):

PR #540, opened 2026-08-24
fix: validate certificate results QC phase

andrewnguyen22:
Changelog:

- Reject certificate-result QCs unless they are precommit votes.
- Recheck the phase before applying certificate results.
- Require post-election proposer messages and their QCs to reference the same view.
- Use the existing x/ namespace consistently for SMT writes, historical proofs, and compaction.
- Harden store bounds, indexing, backups, derived stores, and transactions; reject mismatched transaction message names.

### DIFF
--- a/fsm/state.go
+++ b/fsm/state.go
@@ -15,6 +15,8 @@ import (
 const (
 	CurrentProtocolVersion         = 2
 	slowApplyTransactionsThreshold = 2 * time.Second
+	// defaultPluginStateReadLimit bounds plugin range reads when no explicit limit is provided.
+	defaultPluginStateReadLimit uint64 = 5000
 )
 
 /* This is the 'main' file of the state machine store, with the structure definition and other high level operations */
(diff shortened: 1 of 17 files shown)


fsm/state.go
```diff
@@ -846,9 +848,9 @@ func (s *StateMachine) StateRead(request *lib.PluginStateReadRequest) (response
 		}
 		// calculate entries
 		var entries []*lib.PluginStateEntry
-		// allow 0 limit
+		// apply the default range limit
 		if r.Limit == 0 {
-			r.Limit = math.MaxUint64
+			r.Limit = 5000
```
rem1niscence: I really dont like "magic numbers", please change it to a variable at the top of the file with a quick explanation of what it is

rem1niscence (approved):
just one simple small change, but some of the store changes I think could lead to some sync issues but I will test by running the code and might update this if I find something

rem1niscence:
> just one simple small change, but some of the store changes I think could lead to some sync issues but I will test by running the code and might update this if I find something

nevermind just checked and all good, LGTM!

fsm/state.go
andrewnguyen22: Addressed in f8484c4f: replaced the magic number with the documented top-level defaultPluginStateReadLimit constant.

PR #615, opened 2026-09-16
fix(explorer): restore copy buttons on HTTP origins

moatus:
Explorer copy buttons throw when opened over HTTP on a LAN address because `navigator.clipboard` is unavailable. Route transaction, block, account, validator, order, search, and shared table copy handlers through one small clipboard helper that uses the Clipboard API when available and a textarea fallback otherwise.

Await successful copying before showing confirmation, report failures, and clean up the fallback textarea and restore focus. The fallback uses an editable textarea with a DOM Range and `setSelectionRange` for iOS compatibility. This also consolidates the existing fallback implementations. No new dependencies.

Validation (local only):

- `npm run test:clipboard` — 8 regression tests pass, covering the secure API, HTTP fallback, missing API, rejected permissions, fallback failures/cleanup, explicit iOS-style selection, and an unavailable window selection.
- `npm run build` — passes.
- Local Chromium against the local Canopy node — 26 native clipboard comparisons pass across all affected views, including full hashes/addresses, secure localhost, HTTP with no Clipboard API, and error feedback. The local node had no orders, so the order page used a local response fixture. No uncaught page errors.
- Local WebKit with iPhone emulation — native copy/paste of the full transaction hash passes, with explicit selection, success feedback, and textarea cleanup verified. This is Linux WebKit emulation, not a physical iOS device test.
- `npm run lint` — the same 5 existing errors and 1 warning as unmodified `origin/development`; no new lint findings.

The proposed changelog file was removed as requested in review. Targets `development` as required by CONTRIBUTING.md.


.docs/CHANGELOG.md
pablocampogo, line 1: Please delete this file, it will be difficult to keep up to date

cmd/rpc/web/explorer/src/lib/clipboard.ts
```diff
@@ -0,0 +1,30 @@
...
+    // The Clipboard API is unavailable on HTTP LAN addresses.
+    const textarea = document.createElement('textarea')
+    const activeElement = document.activeElement
+    textarea.value = text
+    textarea.setAttribute('readonly', '')
+    textarea.style.position = 'fixed'
+    textarea.style.opacity = '0'
+    document.body.appendChild(textarea)
+    try {
+        textarea.select()
```
pablocampogo, line 20: On iOS, a readonly textarea plus .select() does not produce a valid selection for execCommand('copy'). The known workaround is to make it contenteditable/non-readonly and use a Range + selection.setSelectionRange(0, text.length). 

Added this since this PR is mainly for Apple users

pablocampogo (changes requested):

pablocampogo (approved):

